PR #1, opened 2021-10-28
Added a Vagrantfile to boot up a development VM

eldaraharonson:
The Vagrantfile boots up a Fedora 32 VM.
File sharing between local machine and VM is set up.
Memory of VM is set to 1024 MB.
Added the '.vagrant' file to the '.gitignore' file


ifireball (approved):
LGTM for now.

PR #7, opened 2021-11-01
django automation

yaelerdr:
Added a script for the VM to run - setup.sh.
Changed VagarantFile so it will run the setup.sh,and changed port to 8000
Added Pipefile and Pipefile.lock

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,0 +1,18 @@
+#!/bin/bash -ex
+# The -e option would make our script exit with an error if any command
+# fails while the -x option makes verbosely it output what it does
+
+# Install Pipenv, the -n option makes sudo fail instead of asking for a
+# password if we don't have sufficient privileges to run it
+sudo -n dnf install -y pipenv
+
+cd /vagrant
+# Install dependencies with Pipenv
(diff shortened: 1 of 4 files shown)


benmenachem (approved):

eldaraharonson (approved):
Looks good!

setup.sh
```diff
@@ -0,0 +1,18 @@
...
+# Install Pipenv, the -n option makes sudo fail instead of asking for a
+# password if we don't have sufficient privileges to run it
+sudo -n dnf install -y pipenv
+
+cd /vagrant
+# Install dependencies with Pipenv
+pipenv sync --dev
+
+# Run database migrations
+pipenv run python manage.py migrate
```
guy9050, line 14: why is `manage.py` not included in the PR? if we merge it like that, it will cause failures

guy9050:
Seems like some file are missing/ were not added to the PR (settings.py, mange.py.. etc)

setup.sh
yaelerdr: Will do

PR #13, opened 2021-11-04
creat a django project with one static web page.

YarinBou:
- basic job board page.


ifireball:
Why are there beyond_tutorial files in this PR? Also it seems that you've checked-in your IDE configuration, not sure youre team wants that in the project repo and it certainly should not be mixed with a functional change.

Finally, crating a Django project involves running scripts that create some files and them making modifications to add your own functionality. You should have auto-generated and manually-written code in different commits, preferably different PRs.

PR #20, opened 2021-11-12
Models creation

guyy13:
Created the following models: Student, Company, Recruiter, EducationalInstitutes.
Also updated the .gitignore file to ignore Pycharm files.

Worked on it together with @benmenachem

### DIFF
--- a/jobboard/models.py
+++ b/jobboard/models.py
@@ -1,0 +1,52 @@
+from django.db import models
+from django.conf import settings
+
+
+class Gender(models.TextChoices):
+    MALE = "M", "Male"
+    FEMALE = "F", "Female"
+    UNSPECIFIED = "U", "Unspecified"
+
+
+class EducationalInstitutes(models.Model):
+    institute_name = models.CharField(max_length=255)
+
+    def __str__(self):
+        return self.institute_name
+
+
+class Company(models.Model):
+    company_name = models.CharField(max_length=255)
+    company_description = models.TextField(max_length=500)
+    company_website_url = models.CharField(max_length=255)
+
+    def __str__(self):
+        return self.company_name
+
+
+class Recruiter(models.Model):
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+    company_id = models.ForeignKey(Company, on_delete=models.RESTRICT, related_name='recruiters')
+    email = models.EmailField(max_length=255)
+    gender = models.CharField(max_length=1, choices=Gender.choices, default=Gender.UNSPECIFIED)
+    phone_number = models.CharField(max_length=10, blank=True, null=True)
+
+    def __str__(self):
+        return f'{self.first_name} {self.last_name}'
+
+
+class Student(models.Model):
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+    email = models.EmailField(max_length=255)
+    date_of_birth = models.DateField()
+    gender = models.CharField(max_length=1, choices=Gender.choices, default=Gender.UNSPECIFIED)
+    phone_number = models.CharField(max_length=10, blank=True, null=True)
+    educational_institute_name = models.ManyToManyField(EducationalInstitutes, related_name='students')
+    graduation_date = models.DateField()
+
+    def __str__(self):
+        return f'{self.first_name} {self.last_name}'
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ dmypy.json
 
 # Vagrant
 .vagrant
+
+# Pycharm
+.idea


jobboard/migrations/0003_alter_recruiter_company_id.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.9 on 2021-11-12 15:23
```
ifireball: It seems that you've created the models, made the migrations, and them made changes and created another migration for those changes.

Since everything ends up going in as a single PR - its better to clean this up and have a single, clean migration file.

ifireball (approved):
I think this PR has too much in it - it mixes, new models, admin, hooking things up to automation and some minor changes like `.gitignore`. Since the overall amount of line is not huge, I can approve this, please please try to aim for cleaner PRs in the future.

In general a PR should usually do one and only one thing.

jobboard/models.py
```diff
@@ -0,0 +1,32 @@
...
+    gender = models.CharField(max_length=1, choices=Gender.choices, default=Gender.UNSPECIFIED)
+    contact_number = models.IntegerField()
+
+
+class Student(models.Model):
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
+    email = models.EmailField(max_length=255)
+    date_of_birth = models.DateField()
+    gender = models.CharField(max_length=1, choices=Gender.choices, default=Gender.UNSPECIFIED)
+    contact_number = models.IntegerField()
```
ifireball: Is this a phone number? 

https://github.com/google/libphonenumber/blob/master/FALSEHOODS.md

jobboard/models.py
```diff
@@ -0,0 +1,32 @@
...
+    institute_university_name = models.CharField(max_length=255)
```
ifireball: Wouldn't it make sense to have a selectable list of universities? Do we expect each student to always belong to a single university? What about 2nd degree students? Or student that did more then one 1st degree? Or students that transferred?

jobboard/models.py
guyy13: Got it, I'll change it to a CharField as I think using a 3rd party library is a bit overdoing it as we're not really going to validate the phone numbers anyway.
Also I'll make it optional (allow blank\null)

ifireball: I did not advocate for the library - its just to make you think about the assumptions that you might be making when you choose a particular type.

yaelerdr (commented):
@ifireball Not sure - Shouldn't we change anything in the settings.py? maybe in the INSTALLED_APPS? or only after we finish our models classes, and split it into different apps..?

ifireball:
> @ifireball Not sure - Shouldn't we change anything in the settings.py? maybe in the INSTALLED_APPS? or only after we finish our models classes, and split it into different apps..?

INTALLED_APPS is for apps not models. However if you plan to split your project into separate apps you better do this before creating the models (As part of your work division design). Moving models between apps is not easy - its like creating a new model and deleting the old, and then you also need to move all the data and code that uses it.

guyy13:
Updated the PR to only include models creations.

benmenachem (approved):

eldaraharonson (approved):

yaelerdr (approved):

ifireball (changes requested):
Model classes and the corresponding migration files should generally go in the same PR - otherwise code that relies on them cannot really work.
Also, how did you test that the code in this PR works? I expect the PR to reflect that (Hint: I suspect you didn't, given my comment above).

guyy13:
> Model classes and the corresponding migration files should generally go in the same PR - otherwise code that relies on them cannot really work. Also, how did you test that the code in this PR works? I expect the PR to reflect that (Hint: I suspect you didn't, given my comment above).

I did test it, I had everything on the same branch, and once I made sure everything is working I used  `git reset --soft upstream/main` to go back and split everything into different PRs.
Anyway, I only did that because I thought you said the original PR had too much in it. If that's not how you meant to split the PRs can you please clarify exactly how you want them to be split? Because I don't think I understand..

YarinBou (approved):

ifireball:
> 
> I did test it, I had everything on the same branch, and once I made sure everything is working I used `git reset --soft upstream/main` to go back and split everything into different PRs. Anyway, I only did that because I thought you said the original PR had too much in it. If that's not how you meant to split the PRs can you please clarify exactly how you want them to be split? Because I don't think I understand..

While the original PR did have too much in it - you shouldn't have split it the the point where you have PRs that don't work individually. A code in a PR *should work independently*, if more code is needed to make it work - that code needs to be in the same PR or already merged.

There were other ways you could've split that PR, I think I've suggested some of them. One example was to keep administrative stuff like `.gitignore` out of functional PRs. Another is to split along application area lines. Who said students and recruiters need to be in the same PR? They could be in different PRs, different APPs and even handled by different engineers.

guyy13:
> > I did test it, I had everything on the same branch, and once I made sure everything is working I used `git reset --soft upstream/main` to go back and split everything into different PRs. Anyway, I only did that because I thought you said the original PR had too much in it. If that's not how you meant to split the PRs can you please clarify exactly how you want them to be split? Because I don't think I understand..
> 
> While the original PR did have too much in it - you shouldn't have split it the the point where you have PRs that don't work individually. A code in a PR _should work independently_, if more code is needed to make it work - that code needs to be in the same PR or already merged.
> 
> There were other ways you could've split that PR, I think I've suggested some of them. One example was to keep administrative stuff like `.gitignore` out of functional PRs. Another is to split along application area lines. Who said students and recruiters need to be in the same PR? They could be in different PRs, different APPs and even handled by different engineers.

Ok, I think I understand what you mean now, since this is already done, is it ok if I make it all into one PR just this time so we can move on?

PR #29, opened 2021-11-16
Create student models and migrations

guyy13:
Created the following student models: Student, EducationalInstitution.
Created migration files to initialize the models and populate the EducationalInstitution table.
Registered the created models on the admin site.


student/models.py
```diff
@@ -1,3 +1,30 @@
-# from django.db import models
+from django.db import models
+from django.conf import settings
 
-# Create your models here.
+
+class Gender(models.TextChoices):
```
ifireball: Given how loaded gender issues are theses days, why do we even need to know it? Let alone store it in the database?

student/models.py
```diff
@@ -1,3 +1,30 @@
...
+class EducationalInstitution(models.Model):
+    institution_name = models.CharField(max_length=255)
+
+    def __str__(self):
+        return self.institution_name
+
+
+class Student(models.Model):
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
+    first_name = models.CharField(max_length=255)
```
ifireball: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

student/models.py
```diff
@@ -1,3 +1,30 @@
...
+    last_name = models.CharField(max_length=255)
+    email = models.EmailField(max_length=255)
+    date_of_birth = models.DateField()
+    gender = models.CharField(max_length=1, choices=Gender.choices, default=Gender.UNSPECIFIED)
+    phone_number = models.CharField(max_length=10, blank=True, null=True)
```
ifireball: `len("+972-54-1234567") > 10`

student/models.py
```diff
@@ -1,3 +1,30 @@
...
+    educational_institution_name = models.ManyToManyField(EducationalInstitution, related_name='students')
```
ifireball: Bad field name - its not the name of the institute - its a link to multiple records (Did you really mean to put an m-to-m relatoion here? Maybe you meant one-to-m?).

student/models.py
```diff
@@ -1,3 +1,30 @@
...
-# Create your models here.
+
+class Gender(models.TextChoices):
+    MALE = "M", "Male"
+    FEMALE = "F", "Female"
+    UNSPECIFIED = "U", "Unspecified"
+
+
+class EducationalInstitution(models.Model):
+    institution_name = models.CharField(max_length=255)
```
ifireball: Are you missing a link field here back to the students model?

student/admin.py
```diff
@@ -1,3 +1,10 @@
-# from django.contrib import admin
+from django.contrib import admin
+from django.apps import apps
 
-# Register your models here.
+# Register your models here. # Register your models here.
+models = apps.get_models()
```
ifireball: it takes 2 lines to register the models you have and 6 to write this loop which I'm not sure you understand. I know Sim showed it, but I think having simpler code here is better.

student/migrations/0002_institute_names.py
```diff
@@ -0,0 +1,35 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('student', '0001_initial'),
+    ]
+
+    def generate_data(apps, schema_editor):
+        from student.models import EducationalInstitution
+
+        institution_names = [
```
ifireball, line 12: Having one name per-line here would probably make the code more readable.

ifireball (changes requested):

student/models.py
ifireball: Also having both `blank=True` and `null=True` seldom makes sense.

student/models.py
guyy13: @ifireball 
I wanted a many to many relation so every institution can access the list of students that are enrolled to it, as well as every student can be enrolled to multiple institutions.. From what I understood if I made a one to many relation it wouldn't allow an institution to be related to multiple students? Or am I missing something?

student/models.py
guyy13: @ifireball 
I thought if i used a many to many field I don't need an additional field here because I can use the related_name of the field in Student.. should I still add a field?

student/models.py
guyy13: Will remove it..

student/models.py
guyy13: I'll change the max length to 15 and leave only blank as true

student/migrations/0002_institute_names.py
guyy13: Alright

student/admin.py
guyy13: @ifireball Can you check it now?

student/models.py
guyy13: Read it, still think sticking with a CharField will be alright.. if you think otherwise lmk..

YarinBou (approved):

student/models.py
eldaraharonson: Maybe we can change it to "current_institution" or something like that since it is a student jobs site so the relevant information on the student would be where he is studying now? And make it a one to many relation for the institutions table. I feel like where he studied in the past is less relevant and can be in his CV.

student/models.py
ifireball: My comment was with regards to the separation between first and last name, why do you need to separate them?

student/models.py
ifireball: I tend to agree with Eldar here, it seems to me that if you link to many institutions you need extra data to explain each link (You studied there in the past and graduated, etc.), but linking to one that is the one you're currently actively enrolled to seems to tell you the full story.

One thing you should consider there - is it possible for a user of the system not to be enrolled at all?

student/models.py
ifireball: You typically need link fields in both directions.

ifireball (changes requested):

student/models.py
guyy13: I don't think it should be possible for a user to not be enrolled at all as the website is intended for students.
I'll create two fields, one for the current place the student is enrolled at and one for any previous places.
So I guess the current place should be a Foreign key field and the previous places should be many to many?

ifireball: I seem to remember you or some other members of your team did not want to go towards a fully structured CV information, so why do you need this for academic institutes?

Everything you put in the model goes towards providing some application feature, what are the features that would be based on the fields that you propose?

eldaraharonson (approved):

student/models.py
ifireball: Not seeing updates in the code here yet

student/models.py
ifireball: Not seeing updates in the code here yet

ifireball (changes requested):

student/models.py
guyy13: We changed the field in Student to a foreign key so I don't think a link back is required?

ifireball: its useful for all kinds of links - suppose you have an institution object in the code - how do you map it back to the students in that institution?

yaelerdr (approved):

ifireball (approved):

PR #32, opened 2021-11-16
Installation of pytest & basic tests

benmenachem:
pytest installation & basic tests codes to make sure purest works well


guyy13 (approved):

YarinBou (approved):

eldaraharonson (approved):

guy9050 (approved):

yaelerdr (approved):

ifireball (changes requested):
I think its best to combine this PR with so basic test code and GitHub actions configuration so we know everything actually works together

guy9050:
Since this PR is no longer just about installing pytest, please update the headline & description

benmenachem:
> Since this PR is no longer just about installing pytest, please update the headline & description

@guy9050 updated the PR name & description

dummy_tests.py
```diff
@@ -0,0 +1,15 @@
+import pytest
```
ifireball: lets put this in the tests.py file of ond of the apps so we know those work - and perhaps put in a real test instead of a dummy one?

ifireball (changes requested):

dummy_tests.py
guyy13: @ifireball 
We can move the test to one of the apps' tests.py file, but I think it's better to stick with a dummy test for now because we don't really have any model functions currently, so there isn't anything we can test (correct me if I'm wrong), and I don't think adding a new model function is really related to this PR.

ifireball: I could argue that having a pure "installation" PR is a bit wasteful - better implement a small piece of functionality and the test for it and add what you need to run the test along the way.

But yeah moving the test can be enough, you know you'll remove it later anyway, and this way you can show me that you made sure the app test files work.

guyy13: Alright, I moved the test already, can we merge it then?

ifireball (approved):

PR #35, opened 2021-11-22
Jobboard models

eldaraharonson:
Added models to the jobboard app.


jobboard/migrations/0002_fixed_data.py
```diff
@@ -0,0 +1,37 @@
...
+            "South",
+            "North",
+            "Tel Aviv",
+            "Haifa",
+            "Jerusalem"
+        ]
+
+        keywords = [
+            "software",
+            "engineer",
```
guy9050: isn't it better to have it as a one keyword?

jobboard/migrations/0003_job_work_from.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('jobboard', '0002_fixed_data'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
guyy13: I think it's better to create a new migration file that creates the table from scratch with this field instead of having a separate file just to add it

jobboard/migrations/0002_fixed_data.py
eldaraharonson: I don't think I understand what you mean

benmenachem (approved):

YarinBou (approved):
Great job Eldar.

guyy13 (approved):
lgtm

yaelerdr (approved):

jobboard/admin.py
```diff
@@ -1,3 +1,5 @@
-# from django.contrib import admin
+from django.contrib import admin
+from django.apps import apps
 
-# Register your models here.
+for model in apps.get_app_config("jobboard").get_models():
```
ifireball: this seems simpler then the code Sim gave you, but I still think that importing the individual models and registering each one would be simpler to understand in the long run, even if it does require slightly more maintainance

jobboard/models.py
```diff
@@ -0,0 +1,53 @@
...
+    city = models.ForeignKey(City, on_delete=models.CASCADE)
+    address = models.CharField(max_length=255)
+    date_created = models.DateField()
+    title_keywords = models.ManyToManyField(JobTitleKeyword)
+
+    def __str__(self):
+        return f"{self.title}, {self.company.name}"
+
+
+class Application(models.Model):
```
ifireball, line 47: since the jobs models seem complex enough, perhaps moving applications to their own app would be worthwhile?

jobboard/migrations/0002_add_fixed_data.py
```diff
@@ -0,0 +1,42 @@
...
+            for name in region_names:
+                Region(name=name).save()
+
+            City(name="Tel Aviv-Yafo", region=Region.objects.get(name="Tel Aviv")).save()
+            City(name="Beersheba", region=Region.objects.get(name="South")).save()
+            City(name="Rishon Leziyyon", region=Region.objects.get(name="Central")).save()
+            City(name="Herziliyya", region=Region.objects.get(name="Tel Aviv")).save()
+
+            for word in keywords:
+                JobTitleKeyword(keyword=word).save()
```
ifireball: What about some example Job records?

ifireball (approved):
Its ok - but I did write down some changes you may want to consider.

jobboard/admin.py
YarinBou: We used this for loop technique for all models.py placement
In admin.py, also for the PR's that have already been approved. Is it bad practice? Or your personal preference?

ifireball: If I shown you this code without explaining what it does - how long would it take you to figure it out? How certain are you about the stability of the APIs you use there such as `get_app_config` and `get_models`?

I susually prefer a simpler solution until the need for a complex one shows up

YarinBou: Ok , on it.

jobboard/migrations/0002_add_fixed_data.py
eldaraharonson: We were planning to add example jobs when developing tests for functions on the models (we have actually started doing that locally).
We have started adding data to the cities and regions models because that data is real and won't be deleted as we progress.

ifireball: you can add "non-real" data too - when you actually go to production you can add configuration options to you application to know that it is running in production mode and should avoid creating the test data

jobboard/models.py
YarinBou: ok, I opened a new pr with a brand new application app. after this PR will approve I will add the model.py

jobboard/migrations/0002_add_fixed_data.py
eldaraharonson: Understood. I added some example job data as instructed.

ifireball (approved):

PR #36, opened 2021-11-23
Create student model functions and tests

guyy13:
Create the following student models functions (and tests for them):
get_institutions, get_institution_students, get_students, get_students_graduate_after

### DIFF
--- a/student/models.py
+++ b/student/models.py
@@ -8,6 +8,13 @@ class EducationalInstitution(models.Model):
     def __str__(self):
         return self.name
 
+    @classmethod
+    def get_institutions_with_enrolled_students(cls):
+        """
+        Return all institutions that have enrolled students as a QuerySet
+        """
+        return cls.objects.filter(students__isnull=False)
+
 
 class Student(models.Model):
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
@@ -21,3 +28,17 @@ class Student(models.Model):
 
     def __str__(self):
         return f'{self.full_name}'
+
+    @classmethod
+    def get_students_enrolled_at(cls, institutions):
+        """
+        Gets all the students that are enrolled at one of the given institutions
+        """
+        return cls.objects.filter(educational_institution__in=institutions)
+
+    @classmethod
+    def get_students_graduate_after(cls, year):
+        """
+        Gets all the students from the DB that graduate after a given year
+        """
+        return cls.objects.filter(graduation_date__year__gte=year)
--- a/student/tests.py
+++ b/student/tests.py
@@ -1,15 +1,84 @@
 import pytest
+from django.contrib.auth.models import User
+from django.db.models import QuerySet
 
-
-def test_basic_math_operation():
-    assert 10 == 2 * 5
+from student.models import Student, EducationalInstitution
+from datetime import date
 
 
 @pytest.fixture
-def numbers_list():
-    return [1, 2, 3, 4]
+def example_institution_a():
+    institution = EducationalInstitution(name="A")
+    institution.save()
+    return institution
 
 
-def test_list_operation(numbers_list):
-    numbers_list = [x + 1 for x in numbers_list]
-    assert numbers_list == [2, 3, 4, 5]
+@pytest.fixture
+def example_institution_b():
+    institution = EducationalInstitution(name="B")
+    institution.save()
+    return institution
+
+
+@pytest.fixture
+def example_student_a(example_institution_a):
+    example_user = User.objects.create_user("test", "testpassword")
+    example_user.save()
+    student = Student(user=example_user, full_name="test user", email="test@test.com", date_of_birth=date(2020, 7, 13),
+                      phone_number="0000000000", educational_institution=example_institution_a,
+                      graduation_date=date(2020, 7, 13))
+    student.save()
+    return student
+
+
+@pytest.fixture
+def example_student_b(example_institution_b):
+    example_user = User.objects.create_user("test2", "testpassword")
+    example_user.save()
+    student = Student(user=example_user, full_name="test user2", email="test2@test.com",
+                      date_of_birth=date(2023, 7, 13),
+                      phone_number="0000000000", educational_institution=example_institution_b,
+                      graduation_date=date(2023, 7, 13))
+    student.save()
+    return student
+
+
+@pytest.mark.django_db
+def test_get_institutions_with_enrolled_students_returns_institutions_with_enrolled_students_as_query_set(
+        example_student_a, example_institution_a, example_institution_b):
+    institution_query_set = EducationalInstitution.get_institutions_with_enrolled_students()
+    assert isinstance(institution_query_set, QuerySet)
+    assert all(isinstance(inst, EducationalInstitution) for inst in institution_query_set)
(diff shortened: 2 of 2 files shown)


benmenachem (approved):
Good job Guy

YarinBou (approved):

student/models.py
```diff
@@ -8,6 +8,21 @@ class EducationalInstitution(models.Model):
...
+
+    @classmethod
+    def get_institutions(cls):
+        return cls.objects.order_by("name")
+
+    """
+    Gets all the students that are enrolled at the called on institution
+    """
+
+    def get_institution_students(self):
```
ifireball: this does not seem to provide any additional logical functionality beyond what Django already provides - unless you have a specific application use for this list, in which case the method name should reflect that.

student/models.py
```diff
@@ -8,6 +8,21 @@ class EducationalInstitution(models.Model):
     def __str__(self):
         return self.name
 
+    """
+    Gets all the institutions from the DB, ordered by their name
+    """
+
+    @classmethod
+    def get_institutions(cls):
```
ifireball: This function name does not indicate the business/application use of this function - do not write funcitons because you think you're going to need them, have a concrete plan about where and why you are going to use them and name them accordingly.

Are you going to have a screen that will just list all institutions indiscriminately?

student/models.py
```diff
@@ -8,6 +8,21 @@ class EducationalInstitution(models.Model):
     def __str__(self):
         return self.name
 
+    """
```
ifireball: In python you put the function docstrings within the body of the function in the 1st line - that makes them available via the `help()` builtin function, show up in IDEs, etc.

student/tests.py
```diff
@@ -1,15 +1,119 @@
...
-def numbers_list():
-    return [1, 2, 3, 4]
+def create_institution():
+    institution = EducationalInstitution(name="A")
+    institution.save()
+    return institution
+
+
+@pytest.fixture
+def create_student(create_user, create_institution):
```
ifireball: Fixtures are things, therefore fixture function names should typically be nouns and not verbs. Consider some of the code you have below:

`create_student.user.id`

It would make more since if it was like the following right?

`example_student.user.id`

student/tests.py
```diff
@@ -1,15 +1,119 @@
...
+    student = Student(user=create_user, full_name="test user", email="test@test.com", date_of_birth=date(2020, 7, 13),
+                      phone_number="0000000000", educational_institution=create_institution,
+                      graduation_date=date(2020, 7, 13))
+    student.save()
+    return student
+
+
+@pytest.mark.django_db
+def test_get_institutions():
```
ifireball: test names should indicate what you are testing, don't be shy about making them long! For example:

`test_get_institutions_returns_all_institutions_as_a_query_set`

student/tests.py
```diff
@@ -1,15 +1,119 @@
...
+    institution_set = EducationalInstitution.get_institutions()
+    assert list(institution_set.values_list("name")) == [
```
ifireball: is this everything you need to make sure of to be able to write the code that calls this function? Wouldn't the code also need to know that the thing returned from the function is a QuerySet, or at lest has some of its key abilities?

ifireball (changes requested):

student/models.py
guyy13: Recruiters can use this functions to look at students only from specific institutions, maybe it's better to create a similar functions inside the Student model and then filter the students in order to achieve this? This way we could also get a list of institutions instead of only one.

student/models.py
guyy13: I think maybe I'll change it to only return institutions that have enrolled students, this way when recruiters want to filter the students they know which institutions they can choose from

student/models.py
guyy13: Got it, I'm a bit new to python, I'll change it :)

student/tests.py
guyy13: Got it.

student/tests.py
guyy13: OK

student/tests.py
guyy13: Ok, I'll add more assertions to check additional stuff

student/models.py
ifireball: is there some question there that I need to answer somehow?

eldaraharonson (approved):

yaelerdr (approved):

student/models.py
```diff
@@ -8,6 +8,13 @@ class EducationalInstitution(models.Model):
     def __str__(self):
         return self.name
 
+    @classmethod
+    def get_institutions_with_enrolled_students(cls):
+        """
+        Return all institutions that have enrolled students as a QuerySet
```
ifireball, line 14: A function docstring should tell me WHY you wrote it - if the code is well written I can usually tell WHAT it does easily enough from the function name and the code.

Same comment goes for the code below, the function names and documentation are still not good enough to let me understand your application business logic

ifireball (approved):
Approving for now because of time constraints - but function names and explanations of business relevance can still be improved.

student/models.py
guyy13: Understood, once this PR gets merged I'll issue another PR to address this

ifireball: Its customary to use GitHub issues of some other issue tracking system to plan and track future work

guyy13: Will do.

student/tests.py
```diff
@@ -1,15 +1,84 @@
...
+    student = Student(user=example_user, full_name="test user2", email="test2@test.com",
+                      date_of_birth=date(2023, 7, 13),
+                      phone_number="0000000000", educational_institution=example_institution_b,
+                      graduation_date=date(2023, 7, 13))
+    student.save()
+    return student
+
+
+@pytest.mark.django_db
+def test_get_institutions_with_enrolled_students_returns_institutions_with_enrolled_students_as_query_set(
```
guy9050, line 47: The name of the test is "test.. returns students **as a query set**", so in that case only first assertion on line 50 asserts this, but you also test here the student names and the institutions.
So either remove those assertions or rename the test.

student/tests.py
```diff
@@ -1,15 +1,84 @@
...
+        example_student_a, example_institution_a, example_institution_b):
+    institution_query_set = EducationalInstitution.get_institutions_with_enrolled_students()
+    assert isinstance(institution_query_set, QuerySet)
+    assert all(isinstance(inst, EducationalInstitution) for inst in institution_query_set)
+    assert list(institution_query_set.values_list("name")) == [(example_institution_a.name,)]
```
guy9050, line 52: I don't understand this, on line 49 you get a list of all the institutions and here you compare that list to the name of `example_institution_a.name` that you put inside a list?

student/tests.py
guyy13: The name also says it returns institutions with enrolled students, and it checks that as well

student/tests.py
guyy13: I'm checking that the whole list that I got back from the function contains only the institutions it should, comparing the whole list makes sure there aren't any objects in the list that shouldn't be there.
I'm comparing the name field because this is how you guys showed us to do it, instead of comparing the institution objects directly.

guy9050 (approved):

PR #37, opened 2021-11-23
Created functions in the Company class & tests

benmenachem:
Created the following tests in the Company class:

1. get_all_recruiters - a function that returns all recruiters of a specified company
2. get_all_companies - a function that returns all the companies in the website

### DIFF
--- a/recruiter/models.py
+++ b/recruiter/models.py
@@ -10,6 +10,21 @@ class Company(models.Model):
     def __str__(self):
         return self.name
 
+    def get_all_recruiters_of_a_specified_company(self):
+        """
+        returns all recruiters of a specified company
+        Input: a company instance
+        Output: all recruiters of the company
+        """
+        return self.recruiters.all()
+
+    @classmethod
+    def get_all_companies(cls):
+        """
+        returns all existing companies in the website
+        """
+        return cls.objects.order_by("name")
+
 
 class Recruiter(models.Model):
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
--- a/recruiter/tests.py
+++ b/recruiter/tests.py
@@ -1,3 +1,47 @@
-# from django.test import TestCase
+import pytest
+from django.contrib.auth.models import User
+from django.db.models import QuerySet
 
-# Create your tests here.
+from recruiter.models import Company, Recruiter
+
+
+@pytest.fixture
+def example_company():
+    company = Company(name="test_company", description="test_company_description", website_url="test_company_web")
+    company.save()
+    return company
+
+
+@pytest.fixture()
+def example_user():
+    user = User.objects.create_user("test", "testPassword")
+    user.save()
+    return user
+
+
+@pytest.fixture
+def example_recruiter(example_company, example_user):
+    recruiter = Recruiter(user=example_user, name="test_recruiter", company=example_company, email="test@email",
+                          phone_number="test_phone")
+    recruiter.save()
+    return recruiter
+
+
+@pytest.mark.django_db
+def test_get_all_companies_returns_all_companies_as_a_query_set(example_company):
+    companies_set = Company.get_all_companies()
+    assert isinstance(companies_set, QuerySet)
+    assert all(isinstance(company, Company) for company in companies_set)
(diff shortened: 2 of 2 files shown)


eldaraharonson (approved):

guyy13 (approved):

recruiter/tests.py
```diff
@@ -1,3 +1,46 @@
...
+    assert (isinstance(companies_set, QuerySet))
+    assert all(isinstance(company, Company) for company in companies_set)
+    assert list(companies_set.values_list("name", "description", "website_url")) == [
+        (create_company.name, create_company.description, create_company.website_url)]
+
+
+@pytest.mark.django_db
+def test_get_all_recruiters_returns_all_recruiters_as_a_query_set(create_recruiter, create_company):
+    recruiter_set = create_company.get_all_recruiters()
+    assert (isinstance(recruiter_set, QuerySet))
```
guy9050: no need for the outer brackets

YarinBou (approved):

recruiter/tests.py
benmenachem: @guy9050 corrected it. everything else looks fine ?

recruiter/models.py
```diff
@@ -10,6 +10,19 @@ class Company(models.Model):
     def __str__(self):
         return self.name
 
+    def get_all_recruiters(self):
```
ifireball: Why do we need this function? Which role does it fulfill within the application's business logic? the function name and documentation should indicate that.

By the time you are done writing your business logic I should be able to get a full picture of what your application will be like on a logical level - which input data it takes, what outputs it generates and which processes is it doing.

recruiter/tests.py
```diff
@@ -1,3 +1,46 @@
...
+import pytest
+from django.contrib.auth.models import User
+from django.db.models import QuerySet
 
-# Create your tests here.
+from recruiter.models import Company, Recruiter
+
+
+@pytest.fixture
+def create_company():
```
ifireball: see comments I placed on other PRs - fixture names souhld be nouns not verbs.

recruiter/tests.py
```diff
@@ -1,3 +1,46 @@
...
+@pytest.fixture
+def create_recruiter(create_company, example_user):
+    recruiter = Recruiter(user=example_user, name="test_recruiter", company=create_company, email="test@email",
+                          phone_number="test_phone")
+    recruiter.save()
+    return recruiter
+
+
+@pytest.mark.django_db
+def test_get_all_companies_returns_all_companies_as_a_query_set(create_company):
```
ifireball: you are not using the fixture in the test body - if you just always need the test data in the DB - perhaps create it in a migration file?

recruiter/tests.py
```diff
@@ -1,3 +1,46 @@
...
+    companies_set = Company.get_all_companies()
+    assert isinstance(companies_set, QuerySet)
+    assert all(isinstance(company, Company) for company in companies_set)
+    assert list(companies_set.values_list("name", "description", "website_url")) == [
+        (create_company.name, create_company.description, create_company.website_url)]
+
+
+@pytest.mark.django_db
+def test_get_all_recruiters_returns_all_recruiters_as_a_query_set(create_recruiter, create_company):
```
ifireball: same comment as above - if you are passing fixtures to a test as parameters you are supposed to use them - some tools like pylint will actually raise an error for you code if you don't.

ifireball (changes requested):

recruiter/models.py
benmenachem: We want this function In case a student wants to get a list of all recruiters of a specified company
I will change the name to make the function more clear

recruiter/tests.py
benmenachem: Will change the names now

recruiter/tests.py
benmenachem: I don't think I understand what you mean
I do use the fixture in the test body and I do use both fixtures in the other function as well

ifireball: ok, I see it now, but still, why not create the test data in a migration file? having some data there will make developing the screens easier later

YarinBou (approved):

ifireball (approved):

PR #47, opened 2021-11-25
Jobboard functions and tests

eldaraharonson:
1. Removed Application model since we were instructed to put it in a
    seperate app.
2. Added functions to the Job model that fetch jobs based on certain
    criteria regarding the jobs' attributes.
3. Added tests for those functions.


benmenachem (approved):

YarinBou (approved):

guyy13:
Could you try to minimize the use of fixtures and use the example data that's already in the DB instead?

guyy13 (approved):

ifireball:
I'm seeing merged migration files being changed, I cannot safely merge this

ifireball (changes requested):
Do not change merged migration files, create new ones.

guyy13:
> Do not change merged migration files, create new ones.

@ifireball 
I fixed it.
The only migrations file that is modified is the one with the example data. We needed more data for the tests, made sense to add the new data to the already existing file.

ifireball (approved):
While changing test data migrations is also something you shouldn't do, because it can break staging systems if you have any, since you don't have one for now, I'm letting it pass

PR #50, opened 2021-11-28
Added recruiter as a FK to Job

yaelerdr:
- Added recruiter as a FK of Job at jobboard\models.py, so Job will be associated with
  the recruiter who published it
- Created new migration file - 005_job_recruiter.py for that added field


benmenachem:
Yael you have some flake8 issues 
Looks like you have lines that has over 120 chars

yaelerdr:
some changes need to be done in the tests

PR #52, opened 2021-11-29
Add fields to the Student model

### LINKED ISSUE #49: Add fields to the Student model
Add gpa , major and work experience field to the student model

guyy13:
Add the following fields to the Student model:
-major
-about
Also changed the tests' fixtures to work with the new fields.

Fixes #49


eldaraharonson (changes requested):
I think that maybe instead of "Work experience" it should be called "About Me" or "Portfolio". Work experience is a bit too specific and students may have relevant experience (for example personal projects) that isn't considered work experience but can still definitely be relevant information for a potential job.
Also I don't think we should have a GPA field, students without high GPA's may feel insecure and not want to become users, also most CVs I've seen don't include GPA's unless they're high.

benmenachem (approved):
lgtm

yaelerdr (approved):

guyy13:
> I think that maybe instead of "Work experience" it should be called "About Me" or "Portfolio". Work experience is a bit too specific and students may have relevant experience (for example personal projects) that isn't considered work experience but can still definitely be relevant information for a potential job. Also I don't think we should have a GPA field, students without high GPA's may feel insecure and not want to become users, also most CVs I've seen don't include GPA's unless they're high.

Done :)

student/models.py
```diff
@@ -24,6 +24,8 @@ class Student(models.Model):
     phone_number = models.CharField(max_length=15, blank=True)
     educational_institution = models.ForeignKey(EducationalInstitution, on_delete=models.PROTECT, default=None,
                                                 related_name="students")
+    major = models.CharField(max_length=255)
```
ifireball: Wouldn't you want to make major a selectable value as opposed to free text?

YarinBou (approved):

student/models.py
guyy13: Umm, yea I guess I could add that, on it.

student/models.py
```diff
@@ -24,6 +24,8 @@ class Student(models.Model):
...
+    about = models.TextField(blank=True)
```
guy9050, line 42: what is this field (about?)?

guyy13: The major field is to specify which major the student is pursuing (which I will change to a choice field).
The about field is a free text field where the student can write anything about themselves, from work experience to projects they've done.

student/models.py
guyy13: Done, I added a few majors to choose from we can add more in the future if we feel like it, for now, I think it'll serve its purpose

eldaraharonson (approved):

student/models.py
```diff
@@ -16,6 +16,20 @@ def get_institutions_with_enrolled_students(cls):
         return cls.objects.filter(students__isnull=False)
 
 
+class Major(models.TextChoices):
+    UNDECIDED = "U", "Undecided"
+    COMPUTER_SCIENCE = "CS", "Computer Science"
+    SOFTWARE_ENGINEER = "SWE", "Software Engineer"
```
YarinBou, line 22: I think that software engineer is more known as "SE"

dmaizel (approved):

student/models.py
guy9050: @guyy13 please resolve this or comment if you disagree so we can finally merge this PR

yaelerdr: Also think SE would be a better choice:) @guyy13

yaelerdr (changes requested):

student/models.py
guyy13: I disagree
https://acronyms.thefreedictionary.com/software+engineer

PR #53, opened 2021-11-30
job_application functions and tests.

YarinBou:


job_application/models.py
```diff
@@ -4,9 +4,25 @@
...
-    job = models.ForeignKey(Job, on_delete=models.CASCADE, related_name='application_job')
-    student = models.ForeignKey(Student, on_delete=models.CASCADE, related_name='application_student')
+    job = models.ForeignKey(Job, on_delete=models.CASCADE,
+                            related_name='application_job')
+    student = models.ForeignKey(
+        Student, on_delete=models.CASCADE, related_name='application_student')
     date_applied = models.DateField()
 
+    @classmethod
+    def get_Applications_by_job(cls, jobParam):
```
guy9050: Function names should be lowercase

guy9050: also jobParam sounds weird but it's not crucial (why not just `job`?)

job_application/models.py
```diff
@@ -4,9 +4,25 @@
...
+        """
+        Allows students to search for a job based on the city it is located in
```
guy9050: wrong description

job_application/models.py
```diff
@@ -4,9 +4,25 @@
...
+        """
+        return cls.objects.filter(job_id=jobParam.id)
+
+    @classmethod
+    def get_Applications_by_student(cls, studentParam):
+        """
+        Allows students to search for a job based on the city it is located in
```
guy9050: wrong description

job_application/tests.py
```diff
@@ -0,0 +1,57 @@
...
+from jobboard.models import Job
+from student.models import Student, EducationalInstitution
+from django.contrib.auth.models import User
+from job_application.models import Application
+import datetime
+from django.db.models import QuerySet
+
+
+# The tests use the data that is already in the db from the migration files
+@pytest.fixture
```
guy9050, line 11: You already have such fixture for institution_a in the students tests (https://github.com/redhat-beyond/SJMaster/blob/main/student/tests.py#L10) , so instead of creating it here again (duplicated code), you should move it to `conftest.py`.
Same thing applies for all the other fixtures that are already created and used in more than one test file.

job_application/tests.py
```diff
@@ -0,0 +1,57 @@
...
+    assert all(isinstance(application, Application) for application in application_temp_1)
+    assert list(application_temp_1.values_list("job", "student", "date_applied")) == [
+        (job_a.id, student_a.user.id, datetime.date(2021, 7, 15))]
+
+
+@pytest.mark.django_db
+def test_get_all_application_by_a_specified_student_returns_all_applications_as_a_query_set(job_application, student_a):
+    job_a = Job.objects.get(title="job_a")
+    application_temp_2 = Application.get_Applications_by_student(student_a)
+    print(application_temp_2)
```
guy9050: why do you print this?

YarinBou: that just some printing I did to check something and that was left over by mistake. will be fixed.

job_application/models.py
YarinBou: the name will be fixed. 
about the var name I think that:
        return cls.objects.filter(job_id=job.id) 
can be confusing no?

job_application/tests.py
guy9050: It's still there :)

YarinBou: yeah, I don't know how I missed it. just delete it again now.

job_application/models.py
```diff
@@ -4,9 +4,25 @@
...
+        Allows recuriters to search for all of the applications for some job.
+        """
+        return cls.objects.filter(job_id=job.id)
+
+    @classmethod
+    def get_applications_by_student(cls, studentParam):
+        """
+        Allows students to search for all of their job applications.
+        """
+        return cls.objects.filter(student__full_name=studentParam.full_name)
```
guyy13: I don't think filtering the students according to full name is a good idea, this way if we have a few students with the same name it can cause problems. 
I think filtering them according to the user id is better (the auth user id).

guyy13 (changes requested):

job_application/tests.py
```diff
@@ -0,0 +1,59 @@
...
+# The tests use the data that is already in the db from the migration files
+@pytest.fixture
+def test_institution():
+    test_institution = EducationalInstitution(name="test_institution")
+    test_institution.save()
+    return test_institution
+
+
+@pytest.fixture
+def test_student(test_institution):
```
guyy13: Maybe change the fixture name to example_student and example_job_application (As suggested by the mentors on previous PRs).

guyy13 (changes requested):

job_application/tests.py
YarinBou: done that :)

job_application/models.py
YarinBou: just working on it, please take a look in my slack question :)

job_application/models.py
```diff
@@ -4,9 +4,25 @@
...
+        Allows recuriters to search for all of the applications for some job.
+        """
+        return cls.objects.filter(job_id=job.id)
+
+    @classmethod
+    def get_applications_by_student(cls, student):
+        """
+        Allows students to search for all of their job applications.
+        """
+        return cls.objects.filter(student_id=student.user.id)
```
guyy13: Are you sure that's working? I'm not sure if Student has an id field because we overrode the PK in Student.
Anyway even if it does, you're comparing the id of a student to the id of the auth user that's associated with the student, these are not the same ids.
I think changing it to 
`return cls.objects.filter(student_user_id = student.user.id)`
is better.

guyy13 (changes requested):

job_application/models.py
YarinBou: Yes it's working we talked about it in The slack, I send you an image of your method and that's send an error.

guyy13: My bad, the correct syntax is
`return cls.objects.filter(student__user_id=student.user.id)`
Notice the double underscore between student and user.
It's working like that I just tested it.

guyy13 (changes requested):

guyy13 (approved):

job_application/tests.py
guy9050: and yet it's still there. maybe you forgot to do git add/commit?

guy9050:
please squash the commits

YarinBou:
That's weird @guy9050  (the deleting thing) now I'm deleting it again and squashing the commits.

guy9050 (approved):

benmenachem (approved):

job_application/tests.py
```diff
@@ -0,0 +1,61 @@
...
+        job=job_a, student=example_student, date_applied=datetime.date(2021, 7, 15))
+    application_a.save()
+    return application_a
+
+
+@pytest.mark.django_db
+def test_get_all_application_by_a_specified_job_returns_all_applications_as_a_query_set(
+        example_job_application, example_student):
+    job_a = Job.objects.get(title="job_a")
+    application_temp_1 = Application.get_applications_by_job(job_a)
```
guy9050: if you get application**s** (in plural), better to call the variable `applications` or `temp_applications` instead of `application_temp_1`, for the readability of the code when you later do `for application in application_temp_1`

job_application/tests.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.mark.django_db
+def test_get_all_application_by_a_specified_student_returns_all_applications_as_a_query_set(
+        example_job_application, example_student):
+    job_a = Job.objects.get(title="job_a")
+    application_temp_2 = Application.get_applications_by_student(
+        example_student)
+    assert isinstance(application_temp_2, QuerySet)
+    assert all(isinstance(application, Application)
+               for application in application_temp_2)
```
guy9050: see previous comment

job_application/models.py
```diff
@@ -4,9 +4,25 @@
...
+    @classmethod
+    def get_applications_by_student(cls, student):
+        """
+        Allows students to search for all of their job applications.
+        """
+        return cls.objects.filter(student__user_id=student.user.id)
+
     def __str__(self):
-        return f"{self.student.name} applied to {self.job}"
+        return f"{self.student.full_name} applied to {self.job}"
```
yaelerdr, line 28: lgtm

job_application/tests.py
yaelerdr: I agree with @guy9050 . Also, think it's better to change the name of the function to `get_all_applications`..

yaelerdr (approved):

YarinBou:
fixed. :)

eldaraharonson (approved):

guy9050:
please squash the commits

YarinBou:
Done. @guy9050

PR #54, opened 2021-12-02
Add recruiter field, added another func in Jobs, created migration and tests

yaelerdr:
- Added recruiter as FK to jobboard/Job
- Created new migration (saved as 0005)
- Added a new function that returns all jobs posted by a specific recruiter ("get_jobs_by_recruiter_id")
- Added test for that function
- Changes recruiter's test - added recruiters saved instances in where they were matched to the expected result


jobboard/models.py
```diff
@@ -73,7 +77,8 @@ def get_jobs_by_keywords(cls, *keywords_as_string):
         jobs_to_return = set()
         for keyword in keywords_as_string:
             keyword_object = JobTitleKeyword.objects.get(keyword=keyword)
-            jobs_to_return.update(set(cls.objects.filter(title_keywords=keyword_object)))
+            jobs_to_return.update(
```
YarinBou, line 80: lgtm

jobboard/models.py
```diff
@@ -97,5 +102,12 @@ def get_jobs_by_company_name(cls, company_name):
         """
         return set(cls.objects.filter(company=Company.objects.get(name=company_name)))
 
-    def __str__(self):
+    @classmethod
+    def get_jobs_by_recruiter(cls, recruiter_name):
+        """
+        Allows students to search for a job based on the company
```
YarinBou, line 108: I think that a better description will be based on the recruiter's name and maybe even pass the send the recruiter himself as an object and comparing with their id because can be two recruiter's with the same name. you can look at what I did in the job_application model.

jobboard/tests.py
```diff
@@ -16,6 +16,16 @@ def test_get_jobs_by_company_name():
...
+@pytest.mark.django_db
+def test_get_jobs_by_recruiter_name():
+    job_a = Job.objects.get(title="job_a")
+    job_c = Job.objects.get(title="job_c")
+    jobs_by_recruiter_a_added_manually = {job_a, job_c}
+    jobs_by_recruiter_a_name_from_test_function = Job.get_jobs_by_recruiter("a")
+    assert isinstance(jobs_by_recruiter_a_added_manually, set)
+    assert jobs_by_recruiter_a_name_from_test_function == jobs_by_recruiter_a_added_manually
+
+
```
YarinBou, line 28: lgtm

jobboard/models.py
yaelerdr: I agree on that one.
The only problem is that I don't think the function can get a Recruiter instance as an argument..
What do you say of letting it get recruiter_id as an input?

guyy13: It can get a recruiter instance as an argument

guyy13:
We talked about adding a major field to the Job model.
Maybe wait until #52 is merged and then add a major field based on the Major text choices model in Student.

jobboard/migrations/0002_add_example_data.py
```diff
@@ -4,12 +4,14 @@
 
 class Migration(migrations.Migration):
     dependencies = [
-        ('jobboard', '0001_initial'),
+        ('jobboard', '0005_job_recruiter'),
```
guy9050, line 7: what's the reason that you add this dependency?

jobboard/models.py
```diff
@@ -1,38 +1,42 @@
 from django.db import models
-from recruiter.models import Company
+from recruiter.models import Company, Recruiter
 import datetime
 
 
 class Region(models.Model):
     name = models.CharField(max_length=255)
 
-    def __str__(self):
```
guy9050: why do you change this?

yaelerdr: It was auto changed for some reason. I'll change it back to what it was

jobboard/migrations/0002_add_example_data.py
yaelerdr: I added this because 0005 is the migration file I created after adding recruiter as new field. Here, I create job objects with non-null recruiters, so 0005 is a dependency of it

eldaraharonson:
> I think that pytest error might happen because the tests of recruiter are fixture tests, and the tests for job are'nt. As to that, I created an instances of recruiter, and saved it to the db.
> 
> It probably fails because of something with the tests of recruiter(?)
> 
> I have tried to add recruiter/test.py to the dependencies of 0002, but the test fails (without it, it passes) :\

I think you need to consider the new recruiters you created in the recruiter test functions because they functions will probably be accessing it in the class functions.

jobboard/models.py
guy9050: @yaelerdr ?

yaelerdr: Changed that to be a recruiter instance @guy9050

guyy13 (approved):

eldaraharonson (approved):

YarinBou (approved):

guy9050:
please squash the commits

yaelerdr:
> please squash the commits

@guy9050 done :)

benmenachem (approved):

guy9050:
I will merge this PR this time, but basically it's too big - next time please divide it to several PR's.

PR #56, opened 2021-12-05
Student and Recruiter functions to check the role of a user

guyy13:
Student app:
Added a function to check if a user is a Student.

Recruiter app:
Added a function to check if a user is a Recruiter.

### DIFF
--- a/recruiter/models.py
+++ b/recruiter/models.py
@@ -35,3 +35,10 @@ class Recruiter(models.Model):
 
     def __str__(self):
         return f'{self.name}'
+
+    @classmethod
+    def is_recruiter(cls, auth_user_id):
+        """
+        For checking if a currently logged in user is a Recruiter
+        """
+        return cls.objects.filter(user_id=auth_user_id).exists()
--- a/student/models.py
+++ b/student/models.py
@@ -58,3 +58,10 @@ class Student(models.Model):
         For recruiters to use when they want to view students according to their graduation year.
         """
         return cls.objects.filter(graduation_date__year__gte=year)
+
+    @classmethod
+    def is_student(cls, auth_user_id):
+        """
+        For checking if a currently logged in user is a Student
+        """
+        return cls.objects.filter(user_id=auth_user_id).exists()
--- a/recruiter/tests.py
+++ b/recruiter/tests.py
@@ -42,3 +42,11 @@ def test_get_all_recruiters_of_a_specified_company_returns_all_recruiters_as_a_q
         (example_recruiter.user.id, example_recruiter.name, example_recruiter.company.id, example_recruiter.email,
          example_recruiter.phone_number),
     ]
+
+
+@pytest.mark.django_db
+def test_if_user_is_a_recruiter_returns_true_if_the_auth_user_is_associated_with_a_recruiter_account(example_recruiter):
+    example_user_not_recruiter = User.objects.create_user("notrecruiter", "password")
+    example_user_not_recruiter.save()
+    assert Recruiter.is_recruiter(example_user_not_recruiter.id) is False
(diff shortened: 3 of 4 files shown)


yaelerdr (approved):
Looks good

eldaraharonson (approved):

YarinBou (approved):

benmenachem (approved):

recruiter/models.py
```diff
@@ -35,3 +35,13 @@ class Recruiter(models.Model):
 
     def __str__(self):
         return f'{self.name}'
+
+    @classmethod
+    def is_recruiter(cls, auth_user_id):
+        """
+        For checking if a currently logged in user is a Recruiter
+        """
+        user = cls.objects.filter(user_id=auth_user_id)
```
guy9050: blank line between docstring and the code

recruiter/models.py
```diff
@@ -35,3 +35,13 @@ class Recruiter(models.Model):
...
+        if user:
+            return True
+        return False
```
guy9050: I don't understand this check - if it's a user than it's necessarily a recruiter?

student/models.py
```diff
@@ -58,3 +58,13 @@ def get_students_that_graduate_after_specified_year(cls, year):
...
         """
         return cls.objects.filter(graduation_date__year__gte=year)
+
+    @classmethod
+    def is_student(cls, auth_user_id):
+        """
+        For checking if a currently logged in user is a Student
+        """
+        user = cls.objects.filter(user_id=auth_user_id)
+        if user:
```
guy9050: see previous 2 comments

guy9050:
Please squash the commits

recruiter/models.py
guyy13: Will fix

recruiter/models.py
guyy13: @guy9050 
I'm checking if there exists a Recruiter with a PK that matches the given auth user id.
If a Recruiter that matches the search is found it will be stored in the "user" QuerySet.
The statement `if user` checks if the QuerySet is empty, if we go into the if statement it means that it isn't empty, meaning there's a recruiter that matched the search. Otherwise, the query set is empty and we return false.

You can think of the `if user` statement as `if (! user.isEmpty())`

recruiter/models.py
guyy13: Umm I don't think there's a blank line it just looks like that (you can see the line numbers)

guyy13:
> Please squash the commits

I really thought separating it into two commits makes sense here, can you please explain why it should be one commit?

guyy13:
Also, I'll work on fixing the conflicts

guyy13:
Fixed conflicts, squashed the commits, and added comments to the functions to better explain them.

recruiter/models.py
guyy13: I added comments to the functions, hopefully it's clear now

recruiter/models.py
```diff
@@ -35,3 +35,14 @@ class Recruiter(models.Model):
...
+    @classmethod
+    def is_recruiter(cls, auth_user_id):
+        """
+        For checking if a currently logged in user is a Recruiter
+        """
+        # Get the Recruiter instance associated with the given authentication user
+        associated_recruiter_instance = cls.objects.filter(user_id=auth_user_id)
+        if associated_recruiter_instance:  # If true, a matching Recruiter was found
+            return True
+        return False
```
guy9050: I think you can combine these 4 lines into 1 - `return cls.objects.filter(user_id=auth_user_id).exists()`

guyy13: Thanks! done.

guy9050 (approved):

PR #60, opened 2021-12-08
Add student account settings page

guyy13:
Add the option for student users to edit their account settings.
Some more details:
- Each student can only access their own account settings page once they are not logged in, trying to get to the edit student account settings link without being logged in as a student will send you to "dead_end.html"
- Edited the base.html to contain a link to the edit account setting page of a Student (only visible if a student is logged in)
- Added a success page

### DIFF
--- a/student/tests.py
+++ b/student/tests.py
@@ -1,7 +1,7 @@
 import pytest
 from django.contrib.auth.models import User
 from django.db.models import QuerySet
-
+from student.forms import UpdateStudentAccountSettingsForm
 from student.models import Student, EducationalInstitution, Major
 from datetime import date
 
@@ -43,6 +43,19 @@ def example_student_b(example_institution_b):
                       graduation_date=date(2023, 7, 13))
     student.save()
     return student
+
+
+@pytest.fixture
+def example_student_a_data_as_dictionary(example_student_a):
+    example_student_a_data = {"full_name": example_student_a.full_name,
+                              "email": example_student_a.email,
+                              "date_of_birth": example_student_a.date_of_birth,
+                              "phone_number": example_student_a.phone_number,
+                              "educational_institution": example_student_a.educational_institution.id,
+                              "major": str(example_student_a.major),
+                              "about": example_student_a.about,
+                              "graduation_date": example_student_a.graduation_date}
+    return example_student_a_data
 
 
 @pytest.mark.django_db
@@ -105,3 +118,49 @@ def test_get_student_raises_exception_for_non_student_user():
     example_user_not_student.save()
     with pytest.raises(Student.DoesNotExist):
         Student.get_student(example_user_not_student.id)
+
+
+@pytest.mark.django_db
+def test_update_student_account_settings_forms_loads_correctly(example_student_a, example_student_a_data_as_dictionary,
+                                                               client):
+    client.force_login(example_student_a.user)
+    response = client.get("/student/account_settings/")
+    assert response.status_code == 200
+    form = response.context["form"]
+    form_initial_data = response.context["form"].initial
+    assert isinstance(form, UpdateStudentAccountSettingsForm)
+    assert all(form_initial_data[key] == example_student_a_data_as_dictionary[key] for key in form_initial_data)
+
+
+@pytest.mark.django_db
+def test_update_student_account_settings_with_invalid_data(example_student_a, example_student_a_data_as_dictionary,
+                                                           client):
+    invalid_email = "guy"
+    client.force_login(example_student_a.user)
+    example_student_a_data_as_dictionary["email"] = invalid_email
+    response = client.post("/student/account_settings/", data=example_student_a_data_as_dictionary)
+    assert response.status_code == 200
+    form = response.context["form"]
+    assert form.is_valid() is False
+
+
+@pytest.mark.django_db
+def test_update_student_account_settings_with_valid_data(example_student_a, example_student_a_data_as_dictionary,
+                                                         client):
+    new_email = "guyyafe@gmail.com"
+    new_full_name = "guy y"
+    new_about = "I like tests"
+    client.force_login(example_student_a.user)
+    example_student_a_data_as_dictionary["email"] = new_email
+    example_student_a_data_as_dictionary["full_name"] = new_full_name
+    example_student_a_data_as_dictionary["about"] = new_about
+    response = client.post("/student/account_settings/", data=example_student_a_data_as_dictionary)
+    assert response.status_code == 302
+    assert response.url == "/account_update_success/"
+    example_student_a_from_db = Student.objects.get(user_id=example_student_a.user.id)
+    # Test that the new data was updated in the DB
+    assert example_student_a_from_db.email == new_email
+    assert example_student_a_from_db.full_name == new_full_name
+    assert example_student_a_from_db.about == new_about
+    # Test that untouched data hasn't changed
+    assert example_student_a_from_db.educational_institution == example_student_a.educational_institution
--- a/student/views.py
+++ b/student/views.py
@@ -1,3 +1,27 @@
-# from django.shortcuts import render
+from django.shortcuts import render, HttpResponseRedirect
+from .models import Student
+from .forms import UpdateStudentAccountSettingsForm
 
-# Create your views here.
+
+def account_update_success(request):
+    is_student = Student.is_student(request.user.id)
+    return render(request, 'account_settings_update_success.html', {"is_student": is_student})
+
+
+def update_student_account_settings_view(request):
+    context = {}
+    try:
+        student_object = Student.objects.get(user_id=request.user.id)
+    except Student.DoesNotExist:
+        return render(request, "dead_end.html")
+    update_student_form = UpdateStudentAccountSettingsForm(request.POST or None, instance=student_object)
+    if update_student_form.is_valid():
+        update_student_form.save()
+        return HttpResponseRedirect("/account_update_success/")
+    context["form"] = update_student_form
+    add_is_student_to_context(request, context)
+    return render(request, "student_account_settings.html", context)
+
+
+def add_is_student_to_context(request, context):
+    context["is_student"] = Student.is_student(request.user.id)
--- a/student/views.py
+++ b/student/views.py
@@ -1,26 +1,29 @@
-from django.shortcuts import render, HttpResponseRedirect
+from django.shortcuts import render, redirect, get_object_or_404
 from .models import Student
 from .forms import UpdateStudentAccountSettingsForm
 
 
 def account_update_success(request):
-    is_student = Student.is_student(request.user.id)
-    return render(request, 'account_settings_update_success.html', {"is_student": is_student})
+    context = {}
+    add_is_student_to_context(request, context)
+    return render(request, 'account_settings_update_success.html', context)
 
 
 def update_student_account_settings_view(request):
     context = {}
-    try:
-        student_object = Student.objects.get(user_id=request.user.id)
-    except Student.DoesNotExist:
-        return render(request, "dead_end.html")
+    student_object = get_object_or_404(Student, user_id=request.user.id)
     update_student_form = UpdateStudentAccountSettingsForm(request.POST or None, instance=student_object)
-    if update_student_form.is_valid():
-        update_student_form.save()
-        return HttpResponseRedirect("/account_update_success/")
-    context["form"] = update_student_form
     add_is_student_to_context(request, context)
-    return render(request, "student_account_settings.html", context)
+    context["form"] = update_student_form
+    if request.method == 'GET':
+        return render(request, "student_account_settings.html", context)
+
+    elif request.method == "POST" and update_student_form.is_valid():
+        update_student_form.save()
+        return redirect("/account_update_success/", context)
+
+    elif request.method == "POST" and not update_student_form.is_valid():
+        return render(request, "student_account_settings.html", context)
 
 
 def add_is_student_to_context(request, context):
--- a/SJMaster/templates/base.html
+++ b/SJMaster/templates/base.html
@@ -3,52 +3,58 @@
 
 <head>
     <!-- Required meta tags -->
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1"/>
     <!-- Bootstrap CSS -->
-    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3" crossorigin="anonymous" />
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css" rel="stylesheet"
+          integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3" crossorigin="anonymous"/>
     <title>SJMaster</title>
 </head>
 
 <body>
-    <header class="site-header">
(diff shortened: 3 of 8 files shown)


YarinBou (approved):
lgtm

yaelerdr (approved):
good job!! looks good

eldaraharonson (approved):

benmenachem (approved):

student/views.py
```diff
@@ -1,3 +1,25 @@
...
+
+class UpdateStudentSettings(UserPassesTestMixin, UpdateView):
+    model = Student
+    fields = ("full_name", "email", "date_of_birth", "phone_number", "educational_institution", "major", "about",
+              "graduation_date")
+    template_name = 'student_account_settings.html'
+    success_url = "/account_update_success"
+
+    def test_func(self):
+        return self.request.user.id == int(self.kwargs['pk'])
```
guy9050: what do you test here?

student/views.py
```diff
@@ -1,3 +1,25 @@
...
+
+    def get_context_data(self, **kwargs):
```
guy9050: what do you need this for?

student/views.py
```diff
@@ -1,3 +1,25 @@
...
+        context = super().get_context_data(**kwargs)
+        context['is_student'] = True
```
guy9050: why are you setting it?

student/views.py
guyy13: @guy9050 
It's not a test (like a pytest test).
The abstract class 'UserPassesTestMixin' allows you to render a view only if the user that sent the request passes a certain test. the 'test_func' method is an abstract method of this class that has to be implemented when inheriting from it.
It checks here that the ID of the user that sent the request is equal to the ID in the URL that they're trying to get to.
This way you can only edit the account details of your own account, and you cannot access any other accounts by modifying the URL.

student/views.py
guyy13: @guy9050 
I needed to check within the HTML whether the logged in user is a Student or a Recruiter.
I couldn't find a way to call a python function from within the HTML, so the solution I found was to pass the user role in the context dictionary to the HTML page.
In this function, I get the context dictionary that is being sent to the page and add an 'is_student' key, this way I can check from within the HTML whether the logged in user is a Student, and display the correct links in the navigation bar.

student/views.py
guyy13: See my comment above

YarinBou (approved):

student/views.py
guy9050: thanks @guyy13

student/views.py
guy9050: I remember there were functions like `is_recruiter()` or `is_student()`, can't you use those? what do you mean by needed to check it from the HTML?

guyy13: @guy9050 
I want to display different links to the user depending on whether they are a recruiter or a student, so I'm using an 'if' statement in the HTML body. I couldn't find a way to call the is_student() function from within the HTML so this is how I did it. If there's a better solution I'm all up for it.

student/views.py
```diff
@@ -1,3 +1,25 @@
...
+from django.views.generic.edit import UpdateView
 
-# Create your views here.
+
+def account_update_success(request):
+    is_student = Student.is_student(request.user.id)
+    return render(request, 'account_settings_update_success.html', {"is_student": is_student})
+
+
+class UpdateStudentSettings(UserPassesTestMixin, UpdateView):
```
ifireball: Do not use class based views:
https://spookylukey.github.io/django-views-the-right-way/

It generally makes the code harder to understand while providing very little benefit.

student/views.py
guy9050: Please see this message Barak posted about class-based viewes
https://redhat-beyond.slack.com/archives/C02GG3Q59D1/p1638093979130600

student/views.py
guyy13: @ifireball 
Um, ok, I'll try to recreate it using view functions..

guyy13: @ifireball 
Done.

student/views.py
guyy13: Fixed it

student/views.py
```diff
@@ -1,3 +1,23 @@
...
+
+def account_update_success(request):
+    is_student = Student.is_student(request.user.id)
+    return render(request, 'account_settings_update_success.html', {"is_student": is_student})
+
+
+def update_student_account_settings_view(request):
+    context = {}
+    try:
+        obj = Student.objects.get(user_id=request.user.id)
```
guy9050: meaningful variable names? i. e `student_object`

student/views.py
```diff
@@ -1,3 +1,23 @@
...
+    except Student.DoesNotExist:
+        return render(request, "dead_end.html")
+    form = UpdateStudentAccountSettingsForm(request.POST or None, instance=obj)
```
guy9050: update_student_form

student/views.py
```diff
@@ -1,3 +1,23 @@
...
+    if form.is_valid():
+        form.save()
+        return HttpResponseRedirect("/account_update_success/")
+    context["form"] = form
+    context["is_student"] = Student.is_student(request.user.id)
```
guy9050: what is this for?

student/views.py
guyy13: I'll change it

student/views.py
guyy13: Same thing I explained to you on 
https://github.com/redhat-beyond/SJMaster/pull/60#discussion_r767267251

student/views.py
```diff
@@ -1,3 +1,23 @@
...
+    try:
+        student_object = Student.objects.get(user_id=request.user.id)
+    except Student.DoesNotExist:
+        return render(request, "dead_end.html")
+    update_student_form = UpdateStudentAccountSettingsForm(request.POST or None, instance=student_object)
+    if update_student_form.is_valid():
+        update_student_form.save()
+        return HttpResponseRedirect("/account_update_success/")
+    context["form"] = update_student_form
+    context["is_student"] = Student.is_student(request.user.id)
```
guy9050: I'm not sure if that's the ideal way to do it, @ifireball WDYT?

guyy13: @guy9050 
I found a much better way to do it,
I added a related name to the link between user and Student, and user and Recruiter, this way in the HTML I can check directly if the logged in user is a student or a recruiter without having to provide anything from the outside, I'll update it in a few minutes so you can check it out.

ifireball: @guy9050  Could you give me more context? What are you asking me about exactly?

@guyy13 Not sure moving logic into the HTML template is a good idea, at the very least it makes to hard to test. I would look for ways of moving it out of there. In this example, and without knowing a lot about the details, I would say that my expectation would be to have completely different page templates for student and requiters with the view selecting between them. Or perhaps even having different views and routes for each.

guyy13: @ifireball 
we do have different views and pages for each, the discussion here is about linking them to the navigation bar, which has to contain different links depending on who is logged in.
I just pushed the changes, please take a look.

guyy13:
This PR became quite big so I found it right to split the commits, if you think otherwise let me know.

student/views.py
ifireball: I see this is no longer in the code, though I'm not sure how are you handling the navigation without this and I don't have the time to dive deeper right now.

When you have a set of context values that need to go into many pages in order to get embedded in shared UI items such as navigation bars, its usually a good idea to move them into a function that gets called from all the relevant views. That makes it easier to know which context items are relevant for the particular view and which are shared.

guyy13: @ifireball 
Ok, anyway I managed to get around this with a solution that allows you to check from within the template directly whether the logged in user is Student or a Recruiter.
I think it's simpler than creating a function like you suggested. Welcome to take a look when you have time.

ifireball: Ok, I see those "ifs" you have in the template. Here is the key question - how do you test that?

guyy13: @ifireball 
I guess I could recreate the same code in a pytest test or something?
How would I test it if I did it as you suggested (a function that contains the shared context)? I'm not sure I understand

ifireball: If you recreate the code in the test - you are not testing the actual code...

To make it testable - you need to move the logic out of the template - one way to do that is to write different included sub-templates for the parts that need to be different according to logic, and put the name of the part you need to include in the context. This allows you to write a test that ensures you include the right sub-template name in the context in various scenarios.

guyy13: @ifireball 
Ok, I'll try to do that later

guyy13:
@ifireball @guy9050 
I added a function to the student/views.py file to add 'is_student' to the context and called it from the relevant views.
Since `is_student` is a model function of Student that also contains tests I assume it's alright.
Please let me know what you think.

guy9050:
> @ifireball @guy9050 I added a function to the student/views.py file to add 'is_student' to the context and called it from the relevant views. Since `is_student` is a model function of Student that also contains tests I assume it's alright. Please let me know what you think.

I am not sure if that's what Barak meant, let's wait for his approval

guyy13:
Fixed conflicts

guyy13:
Added tests for the update form.

student/tests.py
```diff
@@ -105,3 +105,29 @@ def test_get_student_raises_exception_for_non_student_user():
...
+    if form.is_valid():
+        form.save()
+    assert Student.objects.get(user_id=example_student_a.user.id).email == new_email
+    assert Student.objects.get(user_id=example_student_a.user.id).full_name == new_full_name
+    assert Student.objects.get(user_id=example_student_a.user.id).about == new_about
+    # Test inserting bad email, form should be invalid
+    bad_email = "guy"
+    example_student_a_data_edited["email"] = bad_email
+    form = UpdateStudentAccountSettingsForm(instance=example_student_a, data=example_student_a_data_edited)
+    assert form.is_valid() is False
```
guy9050: This is another test where you insert invalid data, please create another test for this

guyy13: fixed

student/views.py
```diff
@@ -1,3 +1,28 @@
...
+    add_is_student_to_context(request, context)
+    return render(request, 'account_settings_update_success.html', context)
+
+
+def update_student_account_settings_view(request):
+    context = {}
+    try:
+        student_object = Student.objects.get(user_id=request.user.id)
+    except Student.DoesNotExist:
+        return render(request, "dead_end.html")
```
ifireball: When you fail to find information  in a we app, you need to return HTTP 404 error code. Django has a useful `get_object_or_404` function for this:

https://docs.djangoproject.com/en/4.0/topics/http/shortcuts/

student/views.py
```diff
@@ -1,3 +1,28 @@
...
+    update_student_form = UpdateStudentAccountSettingsForm(request.POST or None, instance=student_object)
+    if update_student_form.is_valid():
+        update_student_form.save()
+        return HttpResponseRedirect("/account_update_success/")
```
ifireball: Form handling code typically needs to deal with 3 scenarios:

1. When the form is initially fetched and needs to be populated with initial data from the DB
2. When form data is posted from the browser and its invalid
3. When form data is posted from the browser and its valid

Its not clear how this code deals with those scenarios, especially with the distinction between 1 and 2/3. It seems there is some assumption that the form would be invalid if its initialized from DB data only. This might be causing an unnecessary DB save.

student/views.py
```diff
@@ -1,3 +1,28 @@
...
+    context["form"] = update_student_form
+    add_is_student_to_context(request, context)
+    return render(request, "student_account_settings.html", context)
```
ifireball: Try to be consistent and either user `render()` and `redirect()` or `TemplateResponse` and `HttpResponseRedirect`.

student/tests.py
```diff
@@ -105,3 +118,33 @@ def test_get_student_raises_exception_for_non_student_user():
     example_user_not_student.save()
     with pytest.raises(Student.DoesNotExist):
         Student.get_student(example_user_not_student.id)
+
+
+@pytest.mark.django_db
+def test_update_student_account_settings(example_student_a, example_student_a_data_as_dictionary):
```
ifireball: This is not really testing the user flow - only the form object is being tested and that does not really contain much logic in it.

The right way to test is to consider the various flows through the form view:

1.   When the form is initially fetched, populated with initial data from the DB and displayed.
2.  When form data is posted from the browser and its invalid
3.  When form data is posted from the browser and its valid

And write a test for each that used the `client` fixture to emulate the browser activity.

ifireball (changes requested):
I placed some comments. Most have to do with code style and can be fixed later, but there is a testing gap that must be fixed ASAP.

student/views.py
guyy13: I am not sure I understand what you mean here,
about 1, if I managed to get the student object from the DB, why would the form not manage to load?
About 2, the form checks the data automatically and won't let you submit it if the inserted data is invalid.
I also don't understand why the form would be invalid if it's initialized from DB data only..

ifireball: > if I managed to get the student object from the DB, why would the form not manage to load?

What is testing that your code is actually doing that?

>  the form checks the data automatically and won't let you submit it if the inserted data is invalid.

What checks that the view is actually checking for validity and responding properly tro the user instead of, say, throwing a nasty exception on an invalid save attempt?

What you wrote in your last comment are assumptions based on your understanding of how your code and Django works. Tests should not be making such assumptions. They should test that you get the exact behavior you need given controlled inputs without making too many assumptions on the internal implementation. Is there any guarantee that the implementation even uses proper Django Form and View objects? It might change over time. The only thing that can guarantee it works as needed is a good test.

guyy13: I think you mean that I add these tests in the tests file, but this is the views file so I don't understand if anything else is needed here or not.

student/tests.py
guyy13: Okay, so let's say I'm now using the client fixture and I logged in as a student, then I make a POST request to the form, and I want to verify that the fields of the form fill out properly according to the data of the logged in user. How can I check the data that is inside the form right now without submitting it? I tried to look for it online but no luck so far.

guy9050: @guyy13 you can assert the status code that is returned from the POST request

ifireball: You check the form object in the context. Generally, since there is not real logic in the object - just the definition of its relationship with the model object, its may be enough to test its type to make sure the right one s displayed, but you can also check the the form object is associated with the right model object.

In the case of validation testing, you can inspect the `errors` property on the form object or call the `is_valid()` method.

See the reference for Form objects:
https://docs.djangoproject.com/en/4.0/ref/forms/api/#django.forms.Form

student/views.py
ifireball: Ok with regard to the code clarity - for example its hard to see the difference between the behavior on POST and on GET because there is explicit IF for that

student/tests.py
guyy13: @ifireball 
I managed to get the form from the context of the page and test the type of it as you suggested, but I don't understand how to check that the fields in the form are filled out correctly with the request's user details, can you please clarify?
Thanks.
Here's my code currently:
image

guyy13: Ok nvm, I think I managed to do it using `form.initial`

student/views.py
guyy13: Fixed.

student/tests.py
guyy13: I think I managed to implement the tests, please take a look.

student/tests.py
```diff
@@ -105,3 +118,48 @@ def test_get_student_raises_exception_for_non_student_user():
...
+                                                         client):
+    new_email = "guyyafe@gmail.com"
+    new_full_name = "guy y"
+    new_about = "I like tests"
+    client.force_login(example_student_a.user)
+    example_student_a_data_as_dictionary["email"] = new_email
+    example_student_a_data_as_dictionary["full_name"] = new_full_name
+    example_student_a_data_as_dictionary["about"] = new_about
+    response = client.post("/student/account_settings/", data=example_student_a_data_as_dictionary)
+    assert response.status_code == 302
```
ifireball, line 158: Missing test for the redirection target - there is an HTTP header that contains it in the response.

ifireball (approved):
Added a small comment, otherwise its looks good enough.

student/tests.py
guyy13: Done!

guy9050:
looks much better now!

PR #63, opened 2021-12-11
Added get methods to Student and Recruiter models

eldaraharonson:
1. For both Student and Recruiter models added a get method that accepts a
   user id as parameter and returns the student/recruiter corresponding to
   that user id as a Student/Recruiter object.
2. Added tests for both these methods.

### DIFF
--- a/student/tests.py
+++ b/student/tests.py
@@ -92,3 +92,16 @@ def test_if_user_is_a_student_returns_true_if_the_auth_user_is_associated_with_a
     example_user_not_student.save()
     assert Student.is_student(example_user_not_student.id) is False
     assert Student.is_student(example_student_a.user.id) is True
+
+
+@pytest.mark.django_db
+def test_get_user_as_student_object(example_student_a):
+    assert isinstance(Student.get_student(example_student_a.user.id), Student)
+
+
+@pytest.mark.django_db
+def test_get_student_raises_exception_for_non_student_user():
+    example_user_not_student = User.objects.create_user("notstudent", "password")
+    example_user_not_student.save()
+    with pytest.raises(Student.DoesNotExist):
+        Student.get_student(example_user_not_student.id)
--- a/recruiter/models.py
+++ b/recruiter/models.py
@@ -42,3 +42,12 @@ class Recruiter(models.Model):
         For checking if a currently logged in user is a Recruiter
         """
         return cls.objects.filter(user_id=auth_user_id).exists()
+
+    @classmethod
+    def get_recruiter(cls, auth_user_id):
+        """
+        Returns the user as a recruiter object.
+        Should only be called after verifying that the user is
+        a recruiter with the is_recruiter method.
+        """
+        return cls.objects.get(user_id=auth_user_id)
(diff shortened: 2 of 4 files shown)


guyy13 (approved):
lgtm

benmenachem:
Looks good Eldar

benmenachem (approved):

YarinBou (approved):
all of those functions looked good, but I didn't understand what are you used them for?

yaelerdr (commented):
the code looks great, I'm just not sure - what these functions are needed for?

eldaraharonson:
> the code looks great, I'm just not sure - what these functions are needed for?

When a user is logged in on the website we can only access him as a User object meaning I can't access fields that are specific for Student or Recruiter (such as institute for student or company for recruiter). This method recieves the user id and if he is indeed a Student (I can first check that with the boolean method "is_student") then I get the user as a Student object and I can now access his specific Student fields such as university.

student/tests.py
```diff
@@ -92,3 +92,12 @@ def test_if_user_is_a_student_returns_true_if_the_auth_user_is_associated_with_a
...
     assert Student.is_student(example_user_not_student.id) is False
     assert Student.is_student(example_student_a.user.id) is True
+
+
+@pytest.mark.django_db
+def test_get_user_as_student_object(example_student_a):
+    example_user_not_student = User.objects.create_user("notstudent", "password")
+    example_user_not_student.save()
+    with pytest.raises(Student.DoesNotExist):
+        Student.get_student(example_user_not_student.id)
```
guy9050, line 107: how come it's not exist it you created it on line 99?

eldaraharonson: I created a user that is specifically **not** a student in line 99. The "Student.DoesNotExist" exception means that there is no **Student** with that user id in the database but it doesn't mean that it doesn't exist as a user.
Example_student_a is a student though because that's how he was created in line 27.

guy9050: okay but why are you creating it if you don't need it for the test (line 103) ?

eldaraharonson: I am using it for a test. There are 2 tests in this function:
1. In line 103 I am testing the get_student() function with a user that **is** a Student instance.
2. In lines 101-102 I am testing the get_student() function with a user that is **not** a Student instance. In this case the function should throw a "Student.DoesNotExist" exception, I am testing that it indeed throws that exception. This is the syntax for testing exceptions according to what I read here https://www.kite.com/python/answers/how-to-assert-an-exception-is-raised-using-pytest-in-python
If for example in line 102 I write "Student.get_student(example_student_a)"  then the test will fail because there **is** such a student therefore an exception is **not** thrown.

eldaraharonson: Should I just remove this test?

guy9050: okay I get it now, so this is kind of a negative testing which is essentially another test. something like `test_get_student_raises_exception_for_non_student_user`

eldaraharonson: OK I changed it into 2 seperate tests as you suggested. Although I must say that I feel like this PR is being dragged on way longer than it should be. I can understand if this was a conversation we have in one day but each day you give me a different comment and at this rate it will be very hard for us to be ready for the demo on Sunday.

guy9050: @eldaraharonson  Thank you for sharing this, sometimes it happens that people find out new things on the second review and so it gets more time for the pr to get merged. It's not intentional obviously but these things happen and it's good that you speak up when it happens.

PR #64, opened 2021-12-11
Different forms for registration student and for recruiter.

YarinBou:
This PR comes after the basic registration that we made. and its goal is to separate between the student and recruiter registration.
fd23bcf3-5b90-4d88-a758-197ae13e48b4
646fd5eb-1bb1-4a53-a4d7-20ca792a950c


benmenachem:
The pages look great Yarin. 
When a recruiter signs up does he have the option to choose an existing company and as well to register a new company as well ?

YarinBou:
for now, it's only an existing company but you have a point.  maybe for the demo that's good functionality and we will do it later for our final project?  or do you think should i do it for the demo?

benmenachem:
> for now, it's only an existing company but you have a point.  maybe for the demo that's good functionality and we will do it later for our final project?  or do you think should i do it for the demo?

@YarinBou i think for now it's ok but if we find a quick solution to make an option to create a new company it's better. I'll be home soon and think of a way to do that as well

YarinBou:
sure, that's because I'm already getting a hard time creating both user and recruiter in the same form, so to add up a company registration in the same form can take me some more time, check this out and we will talk later.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "C:\\Users\\yarin\\.virtualenvs\\SJMaster-w3yjJ9fd\\Scripts\\python.exe"
```
guyy13: I think you pushed this file by mistake?

guyy13 (changes requested):

student/forms.py
```diff
@@ -0,0 +1,36 @@
...
+
+class studentRegistrationForm(UserCreationForm):
+    educationalInstitution_Queryset = EducationalInstitution.objects.all()
+    full_name = forms.CharField(max_length=255)
+    email = forms.EmailField(required=True)
+    date_of_birth = forms.DateField()
+    phone_number = forms.CharField(max_length=15)
+    educational_institution = forms.ModelChoiceField(educationalInstitution_Queryset)
+    major = forms.TypedChoiceField(choices=Major.choices)
+    about = forms.CharField(max_length=255)
```
guyy13: 255 characters for the about field is too short IMO.
You can change it to: `about = forms.CharField(widget=forms.Textarea)`
This way we can allow a lot of text in this field and the registration form will also indicate it (see attached screenshot).
image

guyy13 (changes requested):

student/forms.py
YarinBou: done.

.vscode/settings.json
YarinBou: yes by git  add .

guyy13 (approved):
now lgtm, good job Yarin

student/forms.py
```diff
@@ -0,0 +1,37 @@
...
+            user.save()
+        user2 = Student(user=user,
+                        full_name=self.cleaned_data['full_name'],
+                        date_of_birth=self.cleaned_data['date_of_birth'],
+                        phone_number=self.cleaned_data['phone_number'],
+                        educational_institution=self.cleaned_data['educational_institution'],
+                        major=self.cleaned_data['major'],
+                        about=self.cleaned_data['about'],
+                        graduation_date=self.cleaned_data['graduation_date'])
+        user2.save()
```
guy9050: why are you overriding the default `save()` method?

YarinBou: what do you mean? that is the way I saw online to save your own form In several different sites and tutorials. 
there is a better or different way to do it?

eldaraharonson (approved):

benmenachem (approved):

student/forms.py
guy9050: The `save` method is a built-in method in to save an instance into a particular model (in this case student), you don't need to override it unless you have some special constrains that you want to add. Here it seems that your function saves the user 3 times in the DB. why? 

@ifireball thoughts?

YarinBou: if I understand correctly this technique user saves the data into the user's table, user2 saves that data into the student's table, the code in line 24 is if commit=false, but the second user. save is if the user sent the form correctly and just then save him. you can look at the discussion here?  
https://stackoverflow.com/a/37820531
how would you do in a different @guy9050? maybe a short example will help. :)

ifireball: I think this can be simplified by making a two step registration, where you first sight up for an account and then fill in additional details to register for a particular role. That way you can have simpler forms where each form needs to work only with one model.

recruiter/forms.py
```diff
@@ -0,0 +1,29 @@
+from django import forms
+from django.contrib.auth.forms import UserCreationForm
+from django.contrib.auth.models import User
+from .models import Recruiter, Company
+
+
+class recuiterRegistrationForm(UserCreationForm):
```
ifireball: Please follow the usual Python Pep8 styling rules - class names should begin with upper case characters.
Not sure why flake8 is letting this pass.

ifireball (changes requested):
While this PR contains a significant amount of logic, I'm not seeing even a single test. Given the time, I don't think you can deliver this for the demo.

recruiter/forms.py
YarinBou: will be fixed if this is the convention.

student/forms.py
YarinBou: I saw some tutorials online on how to do this thing of a single form for 2 tables. and that worked good, I don't think that needs to be changed. that decision for making this partition in the registration form came after long thinking in our team meeting.  we want to keep it this way.

YarinBou:
from my first PR about the registration, you're (@ifireball) keep telling me, again and again, you can't deliver it for the demo, even tho everything works in the first PR about it. and everything working now as well.
the last PR about it was approved without tests. I don't see how a "significant amount of logic" is that's why I think the test's request at this point is just you making your point and looking for reasons for this PR will not be approved for the demo. without an actual reason.
all the other teams in the course sent a registration page without any tests and no one made from this simple action such a big deal.

ifireball:
Application logic needs to have tests, this is an integral part if the demo requirements. If other teams did not implement tests for their features, they may not be able to present them.

YarinBou:
not just other teams, also all of our latest PR's as been sent and approved without test's and also all of the PR's that still waiting are without test's that seem a bit odd that just for my registration page's 4 day's before the demo this is the first time
we have been asked to do them.  this is my feeling that you are specifically looking for this functionality to not succeed. 
maybe behind the scenes that is not what you thinking about, but this is the feeling I'm getting from you in this work.

ifireball:
I'm sorry that you feel this way but:

1. It's your responsibility as a team to ensure your code meets the requirements even if we (the reviewers) forget to remind them to you
2. If far easier for me to avoid reviewing stuff and just merge blindly. Would you rather I do that instead of trying to teach you to be a better engineer?

YarinBou:
I rather you teach me to be a better engineer, I think that about the way you choose to teach me that.

recruiter/tests.py
```diff
@@ -20,6 +22,40 @@ def example_recruiter(example_user):
     return recruiter
 
 
+@pytest.fixture
+def bad_recruiter_example():
```
guy9050: this name is confusing as you're not creating any recruiter as the fixture name states. I'd change it to something like `invalid_recruiters_data`.

student/tests.py
```diff
@@ -86,22 +141,47 @@ def test_get_students_graduate_after_returns_students_that_graduate_after_the_gi
...
-    example_user_not_student = User.objects.create_user("notstudent", "password")
+    example_user_not_student = User.objects.create_user(
+        "notstudent", "password")
     example_user_not_student.save()
     with pytest.raises(Student.DoesNotExist):
         Student.get_student(example_user_not_student.id)
+
+
+@pytest.mark.django_db
+def test_sign_up_valid_stuednt(student_data_example):
```
guy9050: student

recruiter/tests.py
```diff
@@ -46,20 +83,58 @@ def test_get_all_recruiters_of_a_specified_company_returns_all_recruiters_as_a_q
...
+
+@pytest.mark.django_db
+def test_sign_up_valid_recruiter(recruiter_example):
+    form = RecuiterRegistrationForm(data=recruiter_example)
+    if form.is_valid():
+        user = form.save()
+        assert User.objects.filter(pk=user.user.id).exists()
+        assert Recruiter.objects.filter(pk=user).exists()
+    else:
+        assert False
```
guy9050, line 115: If you get to this line, the test will fail because the form is invalid (probably because the data you pass to it from the fixture), and not because what you're actually testing which is the sign up process

recruiter/tests.py
```diff
@@ -46,20 +83,58 @@ def test_get_all_recruiters_of_a_specified_company_returns_all_recruiters_as_a_q
...
+
+
+@pytest.mark.django_db
+def test_sign_up_invalid_recruiter(bad_recruiter_example):
+    for user_data in bad_recruiter_example:
+        invalid = False
```
guy9050, line 121: `invalid_recruiter` will be clearer

recruiter/tests.py
YarinBou: will be fixed

student/tests.py
YarinBou: will be fixed

recruiter/tests.py
YarinBou: will be fixed

recruiter/tests.py
YarinBou: the forms can be invalid because of several reasons one of them is the fact that the user will send "bad" data. What I test here is that the Django form is actually working and save my data in the right places.  and that is an important part of the signup process no? @guy9050

login/templates/registration/register.html
```diff
@@ -2,14 +2,19 @@
...
-        {% csrf_token %} {{ register_form|crispy }}
-        <button class="btn btn-primary" type="submit">Register</button>
-    </form>
-    <p class="text-center">
-        If you already have an account, <a href="/login">login</a> instead.
-    </p>
+  <h1>Register</h1>
+  <form method="POST">
+    {% csrf_token %} {{ register_form|crispy }}
+    <p>Please select your Roll:</p>
```
ifireball: Role*

Also why use a form here and not simple links? The complications this adds to the view code seem unnecessary

recruiter/forms.py
```diff
@@ -0,0 +1,29 @@
+from django import forms
+from django.contrib.auth.forms import UserCreationForm
+from django.contrib.auth.models import User
+from .models import Recruiter, Company
+
+
+class RecuiterRegistrationForm(UserCreationForm):
```
ifireball, line 7: Does this still need to inherit from UserCreationForm? Can it me a simpler ModelForm without all the custom overrides now that you made separate screens?

ifireball (changes requested):
Since this is a big PR - I did not look at everything yet - but already placed some important comments so please have a look.

login/templates/registration/register.html
YarinBou: yes, you are right, this page is no longer in use because the registration got splite between the student and recruiter. I will upload a PR without this page and URL link.

recruiter/forms.py
YarinBou: I use the UserCreationForm inherit when I sent the request.POST argument in the views.py file. that's the way I saw how to do it (I can send the youtube link if it's relevant), maybe you know a better or different way to do it?

recruiter/tests.py
```diff
@@ -32,7 +68,8 @@ def test_get_all_companies_returns_all_companies_as_a_query_set():
 
 @pytest.mark.django_db
 def test_get_all_recruiters_of_a_specified_company_returns_all_recruiters_as_a_query_set(example_recruiter):
-    recruiter_set = Company.get_all_recruiters_of_a_specified_company(Company.objects.get(name="example_company_a"))
```
ifireball, line 35: This reshaping of the code makes the PR bigger and makes it harder to find the code that needs to be reviewed. Please avoid that in the future.

recruiter/tests.py
```diff
@@ -46,20 +83,58 @@ def test_get_all_recruiters_of_a_specified_company_returns_all_recruiters_as_a_q
...
-    example_user_not_recruiter = User.objects.create_user("notrecruiter", "password")
+    example_user_not_recruiter = User.objects.create_user(
+        "notrecruiter", "password")
     example_user_not_recruiter.save()
     with pytest.raises(Recruiter.DoesNotExist):
         Recruiter.get_recruiter(example_user_not_recruiter.id)
+
+
+@pytest.mark.django_db
+def test_getting_the_correct_url(rf, admin_user):
```
ifireball: Why use `RequestFactory` here and not `client`?

login/templates/registration/register.html
YarinBou: done

recruiter/forms.py
ifireball: Are you using a single form to both create a user account and register it as a recruiter or two different forms, one for creating a user and the other to set it as a recruiter? 

If its two different forms, then there is no reason for the recruiter registration for to inherit from the user creation form.

recruiter/tests.py
```diff
@@ -46,20 +83,58 @@ def test_get_all_recruiters_of_a_specified_company_returns_all_recruiters_as_a_q
...
+    request = rf.get('/recruiterRegister/')
+    # Remember that when using RequestFactory, the request does not pass
+    # through middleware. If your view expects fields such as request.user
+    # to be set, you need to set them explicitly.
+    # The following line sets request.user to an admin user.
+    request.user = admin_user
+    response = recruiterRegister(request)
+    assert response.status_code == 200
```
ifireball: This test is messy:

1. The use of "tf" makes it harder to figure out what the test is doing and I don't think its necessary.
2. You only assert on the HTTP response code, and not the content. Is the system displaying the right form?
3. The test name does not indicate what is the expected behavior.

recruiter/tests.py
ifireball: I don't think its useful to test the form object without testing the flow through the view - to test a form use `client` to submit data to the view and then check the stored data and returned values.

ifireball (changes requested):

student/tests.py
```diff
@@ -86,22 +141,47 @@ def test_get_students_graduate_after_returns_students_that_graduate_after_the_gi
...
+@pytest.mark.django_db
+def test_sign_up_invalid_student(invalid_student_data):
+    for user_data in invalid_student_data:
+        invalid = False
+        form = StudentRegistrationForm(data=user_data)
+        try:
+            form.save()
+        except ValueError:
+            invalid = True
+        assert invalid
```
ifireball: I'm seeing tests for the for object but not for the views. What guarantee do we have that the views are actually using that object?

The right way to test a for as through the view via `client`. That way you can test both the validation logic provided by the form and the control/routing logic provided by the view.

ifireball (changes requested):
Added more comment - the tests are incomplete.

recruiter/tests.py
YarinBou: Ok

recruiter/tests.py
YarinBou: I saw that in the menu that you sent to the guy a few days ago. but I will change that.

recruiter/tests.py
YarinBou: I will add the client functionality to the test, but I think that a test the check if to form object himself do the right work and save to the correct places, is also relevant.

student/tests.py
YarinBou: will be fixed. I think this is the same comment on both of the text files. (recruiter and student)

YarinBou:
hey, @ifireball is this the kind of test that you want to see?
image

to check if the system displays the right form?

recruiter/forms.py
YarinBou: I'm using a single form for both, create a user account and register it as a recruiter, I'm also testing this form in the tests.

ifireball:
See the tests im PR #60

YarinBou:
> See the tests im PR #60

yes, I do the form check according to this PR.

guy9050:
Please let us know once the issues and conflicts are resolved it's ready for re-review @YarinBou

YarinBou:
> Please let us know once the issues and conflicts are resolved it's ready for re-review @YarinBou

hey, @guyy13 and @ifireball I'm ready for re-review.

benmenachem (approved):
Tests looks good yarin!
One question tho - I'm not really sure about it ,but doesn't it make more sense to put the tests in the login app since they check the registration and login?

YarinBou:
That was make sense if the login was for both in the same place. But - we make two different forms and functionally for those models and that's why we splitting those tests into the right models.

.vscode/settings.json
ifireball: Ok, are you planning to remove it?

recruiter/tests.py
ifireball: This would be true only if the for testing would provide a significantly bigger amount of coverage, which it won't in this case since `test_new_recruiter_account_with_valid_data` tests exactly the same cod path. As it is its just more code you'd need to maintain that does not provide significant value.

Also having that 'if' and `assert false` makes the test look strange and hard to understand. You could have simply added `assert form.is_valid()` before the other assertions.

In general you want as little logic in your tests as possible in order to make then reliable. A good test has the following structure:

1. Setup of values and other input for the test (Could be partially/fully done in fixtures)
2. Call production code (Usually should be 1 line and avoid any complex logic)
3. Inspect results with assertions.

You should aim to not mix those parts as you do here - But again, this strange test is simply because of the somewhat awkward API of the form objects that were not really designed to be tested directly. The test that does this via the client is much more straight-forward.

recruiter/tests.py
```diff
@@ -46,20 +82,68 @@ def test_get_all_recruiters_of_a_specified_company_returns_all_recruiters_as_a_q
...
+    else:
+        assert False
+
+
+@pytest.mark.django_db
+def test_sign_up_invalid_recruiter(invalid_recruiters_data):
+    for user_data in invalid_recruiters_data:
+        invalid = False
+        form = RecuiterRegistrationForm(data=user_data)
+        try:
```
ifireball, line 123: This is not how you test for an exception. That is what `pytest.raises` is for.

Testing this via the client would both be simpler and also verify that the view routes the error messages properly.

recruiter/views.py
```diff
@@ -55,3 +54,22 @@ def form_valid(self, form):
...
         return super(CreateNewJobForm, self).form_valid(form)
+
+
+def recruiterRegister(request):
+    if request.method == 'POST':
+        form = RecuiterRegistrationForm(request.POST)
+        if form.is_valid():
+            form.save()
+            message = "Your account has been created. You can log in now!"
+            messages.success(request, f'{message}')
```
ifireball, line 65: Was it really necessary to use `messages` here? It seems to me that since you have dedicated view for presenting the success message, this could have been a simple static string in the template.

This seems to add unnecessary complexity to the code, not to mention that it needs to be tested. Are you sure the messages are preserved across the redirection call? Redirection essentially terminates the HTTP call while sending some redirection headers to the browser causing it to essentially make a brand new HTTP call to another page. Things do not necessarily get preserved across different HTTP calls.

I'm not saying this does not work, you probably at least tested this manually but it does something non trivial that:

1. Forces reviewers to look it up, since in order to work it would need to employ some non-trivial tricks like server-based sessions and cookies behind the scenes, and its not safe to just assume that the `messages` module really does all that.
2. Must be tested via automated tests.

Finally, since I looked further down the code, I see that the `recruiter_created_successfully` template does not event attempt to display the messages, so what we essentially have here is a piece of unnecessary code that sends the reviewers into a wild goose chase. Not nice!

student/tests.py
```diff
@@ -164,3 +222,49 @@ def test_update_student_account_settings_with_valid_data(example_student_a, exam
     assert example_student_a_from_db.about == new_about
     # Test that untouched data hasn't changed
     assert example_student_a_from_db.educational_institution == example_student_a.educational_institution
+
+
```
ifireball, line 230: The tests bellow suffer from similar issues to the ones I already mentioned for the recruiter tests.

ifireball (changes requested):
Given the time, I could merge this even with its current code issues, but since PR also contains files that were added unintentionally I cannot.

ifireball (approved):
Merging for now, please open issues to track remaining issues

PR #66, opened 2021-12-14
Jobboard Main Page Display

eldaraharonson:
1. Changed base.html main column to be full size.
2. Added job rendering in jobboard page with respect to the following:
-    If the logged in user is a student then display jobs that correspond to
      the student's major.

image

-    If the logged in user is not a student then display jobs created in the last 6
     months.

image

image

Note that these jobs will be automatically displayed when the user
first logs in. We will later add a search feature to allow for more
thorough job searching.

### DIFF
--- a/jobboard/tests.py
+++ b/jobboard/tests.py
@@ -1,10 +1,32 @@
 import pytest
+from pytest_django.asserts import assertTemplateUsed
 from jobboard.models import Job
 from student.models import Major
-import datetime
+from datetime import date, timedelta
+from django.contrib.auth.models import User
+from student.models import Student, EducationalInstitution
+from recruiter.models import Recruiter
 
 
 # The tests use the data that is already in the db from the migration files
+
+@pytest.fixture
+def example_institution_a():
+    institution = EducationalInstitution(name="A")
+    institution.save()
+    return institution
+
+
+@pytest.fixture
+def example_student_a(example_institution_a):
+    example_user = User.objects.create_user("test", "testpassword")
+    example_user.save()
+    student = Student(user=example_user, full_name="test user", email="test@test.com", date_of_birth=date(2020, 7, 13),
+                      phone_number="0000000000", educational_institution=example_institution_a,
+                      major=Major.COMPUTER_SCIENCE, about="testing", graduation_date=date(2020, 7, 13))
+    student.save()
+    return student
+
 
 @pytest.mark.django_db
 def test_get_jobs_by_company_name():
@@ -83,7 +105,7 @@ def test_get_jobs_posted_on_or_after_specific_date():
     job_d = Job.objects.get(title="job_d")
     jobs_created_from_specific_date_added_manually = {job_b, job_c, job_d}
     jobs_created_from_specific_date_from_function = \
-        Job.get_jobs_posted_on_or_after_specific_date(datetime.date(2021, 7, 15))
+        Job.get_jobs_posted_on_or_after_specific_date(date(2021, 7, 15))
     assert jobs_created_from_specific_date_added_manually == jobs_created_from_specific_date_from_function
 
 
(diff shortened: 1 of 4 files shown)


benmenachem (approved):
Looks great Eldar!

YarinBou (approved):

jobboard/views.py
```diff
@@ -1,5 +1,76 @@
...
+    jobs_by_student_major = Job.get_jobs_by_major(user_as_student.major)
+    content = {'jobs_to_display': jobs_by_student_major,
+               'user': request.user,
+               'user_is_student': True,
+               'user_is_recruiter': False,
+               'user_as_student': user_as_student}
+    return content
+
+
+def get_content_if_user_is_recruiter(request):
```
guyy13: I made a "My Jobs" page for recruiters that displays all the jobs that the logged in recruiter has posted, as well as applicants for each job. As I recall this is what we agreed on in our last meeting.
I think it's better to redirect a recruiter to the "My Jobs" page when they log in instead of the view you created here.

guyy13 (changes requested):

jobboard/views.py
guyy13: And I think we should still allow a recruiter to browse the jobs on the website, so when a recruiter goes to the main page they can be treated as if they're not logged in and see all the jobs from the past 6 months.

jobboard/views.py
```diff
@@ -1,5 +1,76 @@
...
+    jobs_created_by_recruiter = Job.get_jobs_by_recruiter_id(user_as_recruiter)
+    content = {'jobs_to_display': jobs_created_by_recruiter,
+               'user': request.user,
+               'user_is_student': False,
+               'user_is_recruiter': True,
+               'user_as_recruiter': user_as_recruiter}
+    return content
+
+
+def get_date_six_months_ago():
```
guy9050: 1. why do you need this function?
2. I think that your function won't work for some special end cases, i.e for months with 31 days that 6 months earlier don't have 31 days
3. look up `datetime` - it'll be easier to do that with it

guyy13:
One more thing, we wanted each job to link to a different page where you can view the entire job details, and then also have an "Apply" button, are you planning to push that in a different PR or are we skipping it for now?

jobboard/views.py
eldaraharonson: > I made a "My Jobs" page for recruiters that displays all the jobs that the logged in recruiter has posted, as well as applicants for each job. As I recall this is what we agreed on in our last meeting. I think it's better to redirect a recruiter to the "My Jobs" page when they log in instead of the view you created here.

I don't remember that but you're right, I will redirect a recruiter to that page.

eldaraharonson: > And I think we should still allow a recruiter to browse the jobs on the website, so when a recruiter goes to the main page they can be treated as if they're not logged in and see all the jobs from the past 6 months.

And yes that's a good idea. I will make these changes

eldaraharonson:
> One more thing, we wanted each job to link to a different page where you can view the entire job details, and then also have an "Apply" button, are you planning to push that in a different PR or are we skipping it for now?

Yes this will be a different PR. @yaelerdr is doing that one.

jobboard/views.py
eldaraharonson: Yes I see now that some edge cases won't work, I will find a solution.

guyy13 (approved):

guyy13 (changes requested):
Eldar I think you misunderstood me again, there's no need for a "My Jobs" button as it will be linked on the nav bar.
If a recruiter goes to the jobboard page, they should see the latest jobs, just like a user who is not logged in sees them, that's it, nothing else is needed to be done.

yaelerdr (approved):

guy9050:
what about testing it?

eldaraharonson:
> what about testing it?

Added tests now

jobboard/tests.py
```diff
@@ -112,3 +133,48 @@ def test_get_jobs_by_major():
...
+    client.force_login(example_user)
+    response = client.get("/")
+    assert response.status_code == 200
+    assert response.context['user'] == example_user
+    assert response.context['user_is_student'] is False
+    assert response.context['user_is_recruiter'] is False
+    date_six_months_ago = date.today() - timedelta(days=180)
+    jobs_to_display = Job.get_jobs_posted_on_or_after_specific_date(date_six_months_ago)
+    # test that jobs displayed are only jobs created 6 months ago
+    assert response.context['jobs_to_display'] == jobs_to_display
```
ifireball, line 152: Missing assertion that the right template is used, same for the other tests below.

ifireball (approved):
There is missing coverage in the tests, please either fix or open an issue, and then I'll merge this.

jobboard/tests.py
eldaraharonson: I just added the relevant assertions.

jobboard/tests.py
```diff
@@ -112,3 +134,51 @@ def test_get_jobs_by_major():
...
+
+
+@pytest.mark.django_db
+def test_page_display_to_user_that_is_student(client, example_student_a):
+    client.force_login(example_student_a.user)
+    response = client.get("/")
+    assert response.status_code == 200
+    assertTemplateUsed(response, 'jobboard/board.html')
```
ifireball, line 160: This is not the pytest-style of doing this, you could just assert on `response.template_name` instead.

ifireball (approved):
Merging for now, even though some unitest-style assertions were written instead of pytest ones. Please fix this later.

PR #67, opened 2021-12-15
Adding links to navbar

benmenachem:
* Added links from navbar to the recruiter' account settings & create new job form page
* Added the major field to the create new job form

### DIFF
--- a/SJMaster/templates/base.html
+++ b/SJMaster/templates/base.html
@@ -29,31 +29,40 @@
         >
           <span class="navbar-toggler-icon"></span>
         </button>
-
-        <div class="collapse navbar-collapse" id="navbarSupportedContent">
-          <ul class="navbar-nav mr-auto">
-            <li class="nav-item active">
-              <a class="nav-link" href="/">Job Search</a>
-            </li>
-            {% if user.is_authenticated %} {% if is_student %}
-            <li class="nav-item">
-              <a class="nav-link" href="/student/account_settings/"
-                >Account Settings</a
-              >
-            </li>
-            {% endif %}
-            <li class="nav-item">
-              <a class="nav-link" href="/logout">Logout</a>
-            </li>
-            <li class="nav-item">
-              <a class="nav-link" href="#">Welcome, {{ user.username }}</a>
-            </li>
-            {% else %}
-            <li class="nav-item">
-              <a class="nav-link" href="/login">Login</a>
-            </li>
-            {% endif %}
-          </ul>
+            <div class="collapse navbar-collapse" id="navbarSupportedContent">
+                <ul class="navbar-nav mr-auto">
+                    <li class="nav-item active">
+                        <a class="nav-link" href="/">Job Search</a>
+                    </li>
+                    {% if user.is_authenticated %}
+                    {% if is_recruiter %}
+                     <li class="nav-item">
+                        <a class="nav-link" href="recruiter/account_settings/{{ user.id }}">Account Settings</a>
+                    </li>
+                     <li class="nav-item">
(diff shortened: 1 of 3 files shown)


SJMaster/templates/base.html
```diff
@@ -24,9 +24,14 @@
...
                     </li>
                     {% if user.is_authenticated %}
-                    <li class="nav-item">
-                        <a class="nav-link" href="#">Account Settings</a>
+                    {% if is_recruiter %}
+                     <li class="nav-item">
+                        <a class="nav-link" href="recruiter/account_settings/{{ user.id }}">Account Settings</a>
+                    </li>
+                     <li class="nav-item">
+                        <a class="nav-link" href="recruiter/create_new_job_form">Create a new job</a>
```
guyy13: Maybe change it to "Create Job"?

guyy13 (changes requested):

SJMaster/templates/base.html
benmenachem: fine

guyy13 (approved):
lgtm

guyy13 (approved):

yaelerdr (approved):

eldaraharonson (approved):

ifireball (changes requested):
No tests? Am I supposed to guess if this code does what you intended it to do?

benmenachem:
> No tests? Am I supposed to guess if this code does what you intended it to do?

I'll update the PR with tests later today

benmenachem:
> > No tests? Am I supposed to guess if this code does what you intended it to do?
> 
> @ifireball I wasn't really sure about the tests for this PR.
 I consulted with @guyy13 and we agreed that the entire logic of this PR relies on the is_student and is_recruiter model functions that already have tests. therefore I don't think there's anything test here as you instructed not to test the content of the html itself.

benmenachem:
> > No tests? Am I supposed to guess if this code does what you intended it to do?

 @ifireball I wasn't really sure about the tests for this PR. I consulted with @guyy13 and we agreed that the entire logic of this PR relies on the is_student and is_recruiter model functions that already have tests. therefore I don't think there's anything test here as you instructed not to test the content of the html itself.

ifireball:
This PR is not devoid of logic. I think we discussed how to move logic out of templates and into testable locations somewhere already.

Right now there is no test that gives you any form of verification that the right links are displayed under the right conditions.

guyy13:
> This PR is not devoid of logic. I think we discussed how to move logic out of templates and into testable locations somewhere already.
> 
> Right now there is no test that gives you any form of verification that the right links are displayed under the right conditions.

We did discuss it and if I recall correctly you said we'd have to create separate views or even templates in order to move the logic out. Since this is a shared navigation bar across all of the pages I'm not sure how we would do that here. We could try to test the content of the HTML itself in order to check that the links are correct, but again as I recall you said to avoid it. Anyway I really don't understand why we need to test it, we test the context that we pass to the html, if we test that the correct links are showing up we basically test the if statement of Python.

ifireball:
A. Of course you need to test the 'if' statements those commonly get messed up as code is maintained over time.
B. We talked about having different included template files, not different views
C. There is at least one new context updating function in this MR that I don't see tests for either for itself or for the views that use it

guyy13:
> A. Of course you need to test the 'if' statements those commonly get messed up as code is maintained over time.
> B. We talked about having different included template files, not different views
> C. There is at least one new context updating function in this MR that I don't see tests for either for itself or for the views that use it

Ok, we'll try to figure it out

PR #68, opened 2021-12-17
Changed the recruiter's update settings page & added tests

benmenachem:
Since the PR for the recruiter's account settings page was merged before @ifireball said it is better not to use updateView class in order to update forms I changed the page & added tests

### DIFF
--- a/recruiter/tests.py
+++ b/recruiter/tests.py
@@ -2,9 +2,8 @@ import pytest
 from django.contrib.auth.models import User
 from django.db.models import QuerySet
 from pytest_django.asserts import assertTemplateUsed
-
 from recruiter.models import Company, Recruiter
-from recruiter.forms import RecuiterRegistrationForm
+from recruiter.forms import RecuiterRegistrationForm, UpdateRecruiterAccountSettingsForm
 from jobboard.models import Job
 from job_application.models import Application
 from student.models import EducationalInstitution, Student
@@ -20,9 +19,10 @@ def example_user():
 
 @pytest.fixture
 def example_recruiter(example_user):
-    recruiter = Recruiter(user=example_user, name="yarin bouzaglo",
-                          company=Company.objects.get(name="example_company_a"), email="test@yarin@example.com",
-                          phone_number="0524434795")
+    recruiter = Recruiter(user=example_user, name="test_recruiter",
+                          company=Company.objects.get(name="example_company_a"), email="test@email",
+                          phone_number="0543534343")
+
     recruiter.save()
     return recruiter
 
@@ -193,7 +193,7 @@ def test_new_recruiter_account_with_valid_data(valid_recruiter_data, client):
 
 
 @pytest.mark.django_db
-def test_recruiter_my_jobs_loads_correct_data(example_recruiter, example_job_application, client):
+def test_recruiter_my_jobs_loads_correct_data(client):
     recruiter_object = Recruiter.objects.get(name="a")
     client.force_login(recruiter_object.user)
     response = client.get("/myjobs")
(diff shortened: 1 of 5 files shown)


guyy13 (approved):
LGTM

yaelerdr (approved):

eldaraharonson (approved):

YarinBou (approved):

SJMaster/urls.py
```diff
@@ -16,35 +16,29 @@
...
-                             job_created_successfully,
-                             recruiterRegister,
-                             recruiter_created_successfully,
-                             recruiter_view_my_jobs_and_applications,)
+from recruiter.views import (
+    CreateNewJobForm,
+    job_created_successfully,
+    recruiterRegister,
+    recruiter_created_successfully,
+    recruiter_view_my_jobs_and_applications, )
```
guy9050: what's this space for?

recruiter/tests.py
```diff
@@ -14,9 +16,10 @@ def example_user():
...
 @pytest.fixture
 def example_recruiter(example_user):
-    recruiter = Recruiter(user=example_user, name="yarin bouzaglo",
-                          company=Company.objects.get(name="example_company_a"), email="test@yarin@example.com",
-                          phone_number="0524434795")
+    recruiter = Recruiter(user=example_user, name="test_recruiter",
+                          company=Company.objects.get(name="example_company_a"), email="test@email",
+                          phone_number="0543534343")
+
     recruiter.save()
```
guy9050, line 26: since it's a fixture and you only need it for the tests, it doesn't seem right to me that it's saved into the DB. @ifireball thoughts?

SJMaster/urls.py
benmenachem: looks like a typo. fixed it in PR #69 so when it will be merged it'll be fixed

recruiter/tests.py
benmenachem: When we just created the models we used pytest fixtures as well,  and of what I remember each test creates some sort of a temporary data base so the instance isn't saved in the real data base.

SJMaster/urls.py
```diff
@@ -16,35 +16,29 @@
 from django.contrib import admin
 from django.urls import path, include
 from jobboard.views import board
+from recruiter.views import update_recruiter_account_settings_view
```
ifireball: Please try to consolidate this with the import below - so there is only one import line for `recruiter.views`

recruiter/tests.py
```diff
@@ -147,3 +150,59 @@ def test_new_recruiter_account_with_valid_data(valid_recruiter_data, client):
...
+def test_update_recruiter_account_settings_forms_loads_correctly(example_recruiter,
+                                                                 example_recruiter_data_as_dictionary,
+                                                                 client):
+    client.force_login(example_recruiter.user)
+    response = client.get("/recruiter/account_settings/")
+    assert response.status_code == 200
+    form = response.context["form"]
+    form_initial_data = response.context["form"].initial
+    assert isinstance(form, UpdateRecruiterAccountSettingsForm)
+    assert all(form_initial_data[key] == example_recruiter_data_as_dictionary[key] for key in form_initial_data)
```
ifireball, line 242: Missing template check. For the views below as well.

ifireball (changes requested):

SJMaster/urls.py
benmenachem: fixed

recruiter/tests.py
benmenachem: added the requested template checks. please review the PR if you can I think everything is okay now. @ifireball

SJMaster/urls.py
```diff
@@ -17,34 +17,28 @@
...
 from student.views import studentRegister, student_created_successfully
 
-
 urlpatterns = [
     path('admin/', admin.site.urls),
     path('', board, name='board'),
     path('', include("django.contrib.auth.urls"), name="login"),
-    path('recruiter/account_settings/<slug:pk>',
-         UpdateRecruiterSettings.as_view(), name='recruiter_account_settings'),
     path('recruiter/create_new_job_form/',
```
ifireball: Is this line duplicated on line 52 below? Here you can see the danger of moving code around too much in a PR that is meant to add functionality - I could've missed that...

Also it seems a little strange now that we're importing a form object for a "views" module, not to mention using an object here at all as opposed to a plain function.

ifireball (changes requested):

SJMaster/urls.py
benmenachem: @ifireball Your right, there were actually a few duplicated lines in the urls file I just fixed them. I will make sure to fix conflicts from now and on more carefully in order to avoid these kind of mistakes.

recruiter/views.py
```diff
@@ -27,14 +28,32 @@ def recruiter_view_my_jobs_and_applications(request):
...
+
+    elif request.method == "POST" and update_recruiter_form.is_valid():
+        update_recruiter_form.save()
+        return redirect("/account_update_success/", context)
+
+    elif request.method == "POST" and not update_recruiter_form.is_valid():
+        return render(request, "recruiter_account_settings.html", context)
+
+
+def add_is_recruiter_to_context(request, context):
```
guyy13: I think you can delete this function as the "is_recruiter" key is no longer required in the context dict.

benmenachem: fixed

guyy13 (approved):

ifireball (approved):

PR #69, opened 2021-12-17
Updated Create new job form & added tests

benmenachem:
Updated the create new job form to be function based and not based on CreateViews
Also added tests for the function together with @guyy13

### DIFF
--- a/jobboard/tests/conftest.py
+++ b/jobboard/tests/conftest.py
@@ -1,4 +1,7 @@
 import pytest
+import jobboard
+import student
+from recruiter.models import Recruiter, Company
 from student.models import Major
 from datetime import date
 from django.contrib.auth.models import User
@@ -21,3 +24,27 @@ def example_student_a(example_institution_a):
                       major=Major.COMPUTER_SCIENCE, about="testing", graduation_date=date(2020, 7, 13))
     student.save()
     return student
+
+
+@pytest.fixture
+def example_recruiter():
+    example_user = User.objects.create_user("test", "testpassword")
+    example_user.save()
+    recruiter = Recruiter(user=example_user, name="test_recruiter",
+                          company=Company.objects.get(name="example_company_a"), email="test@email",
+                          phone_number="test_phone")
+    recruiter.save()
+    return recruiter
+
+
+@pytest.fixture
+def example_job_data_as_dictionary():
+    example_job = {"title": 'example_job',
+                   "major": student.models.Major.COMPUTER_SCIENCE,
+                   "job_type": 'Full-Time',
+                   "work_from": 'Office-Only',
+                   "description": 'Junior software engineer',
+                   "city": jobboard.models.City.objects.get(name="Tel Aviv-Yafo").id,
+                   "address": 'Hashalom',
+                   "title_keywords": [jobboard.models.JobTitleKeyword.objects.get(keyword="software").id]}
+    return example_job
--- a/recruiter/views.py
+++ b/recruiter/views.py
@@ -1,18 +1,11 @@
 from django.contrib import messages
 from .forms import RecuiterRegistrationForm
-from django.contrib.auth.mixins import UserPassesTestMixin
-from django.shortcuts import render, get_object_or_404, redirect
-from django.views.generic import CreateView
+from django.shortcuts import render, redirect, get_object_or_404
 from jobboard.models import Job
 from recruiter.models import Recruiter
 from job_application.models import Application
-from datetime import date
 import jobboard.views
 from .forms import UpdateRecruiterAccountSettingsForm
-
-
-def job_created_successfully(request):
-    return render(request, 'job_created_successfully.html')
 
 
 def recruiter_view_my_jobs_and_applications(request):
@@ -48,24 +41,6 @@ def update_recruiter_account_settings_view(request):
 
     elif request.method == "POST" and not update_recruiter_form.is_valid():
         return render(request, "recruiter_account_settings.html", context)
-
-
-class CreateNewJobForm(CreateView, UserPassesTestMixin):
-    model = Job
-    fields = (
-        'title', 'job_type', 'work_from', 'description', 'city', 'address', 'title_keywords')
-    template_name = 'create_new_job_form.html'
-    success_url = '/job_created_successfully'
-
-    def test_func(self):
-        return Recruiter.is_recruiter(self.request.user.id)
-
-    def form_valid(self, form):
-        logged_recruiter = Recruiter.objects.get(user_id=self.request.user.id)
-        form.instance.recruiter = logged_recruiter
-        form.instance.company = logged_recruiter.company
-        form.instance.date_created = date.today()
-        return super(CreateNewJobForm, self).form_valid(form)
 
 
 def recruiterRegister(request):
--- a/jobboard/forms.py
+++ b/jobboard/forms.py
@@ -1,0 +1,8 @@
+from django import forms
+from jobboard.models import Job
+
+
+class CreateNewJobForm(forms.ModelForm):
+    class Meta:
+        model = Job
+        fields = ['title', 'job_type', 'major', 'work_from', 'description', 'city', 'address', 'title_keywords']
--- a/jobboard/tests/view_tests.py
+++ b/jobboard/tests/view_tests.py
@@ -1,9 +1,11 @@
 import pytest
 from pytest_django.asserts import assertTemplateUsed
+from jobboard.forms import CreateNewJobForm
 from jobboard.models import Job
 from datetime import date, timedelta
 from django.contrib.auth.models import User
 from recruiter.models import Recruiter
+
 
 # The tests use the data that is already in the db from the migration files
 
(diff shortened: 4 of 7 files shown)


guyy13 (approved):
LGTM

jobboard/tests.py
```diff
@@ -182,3 +186,57 @@ def test_page_display_to_user_that_is_recruiter(client):
...
+
+@pytest.mark.django_db
+def test_create_new_job_form_with_invalid_data(example_recruiter, example_job_data_as_dictionary,
+                                               client):
+    example_job_data_as_dictionary['title'] = ""  # invalid title
+    client.force_login(example_recruiter.user)
+    response = client.post("/recruiter/create_new_job_form/", data=example_job_data_as_dictionary)
+    assert response.status_code == 200
+    form = response.context["form"]
+    assert form.is_valid() is False
```
ifireball: Missing test for the "simple empty form display" code path.

Also there is no test that the form is actually displayed to the end user (e.g. make sure that the right template is used)

eldaraharonson (approved):

yaelerdr (approved):

YarinBou (approved):

SJMaster/urls.py
```diff
@@ -16,26 +16,22 @@
...
 from jobboard.views import board
+from recruiter.views import create_new_job_form
 from student.views import update_student_account_settings_view, account_update_success
 from recruiter.views import (UpdateRecruiterSettings,
-                             CreateNewJobForm,
                              job_created_successfully,
                              recruiterRegister,
                              recruiter_created_successfully,
-                             recruiter_view_my_jobs_and_applications,)
+                             recruiter_view_my_jobs_and_applications, )
```
guy9050: what's this space for?

jobboard/tests.py
```diff
@@ -182,3 +186,57 @@ def test_page_display_to_user_that_is_recruiter(client):
...
+                   "work_from": 'Office-Only',
+                   "description": 'Junior software engineer',
+                   "city": jobboard.models.City.objects.get(name="Tel Aviv-Yafo").id,
+                   "address": 'Hashalom',
+                   "title_keywords": [jobboard.models.JobTitleKeyword.objects.get(keyword="software").id]}
+    return example_job
+
+
+@pytest.mark.django_db
+def test_create_new_job_form_with_valid_data(example_recruiter, example_job_data_as_dictionary,
```
guy9050: Change the parameters in the order that they are used, i.e `client` is used first so it should be the first one

jobboard/tests.py
```diff
@@ -182,3 +186,57 @@ def test_page_display_to_user_that_is_recruiter(client):
...
+        "title_keywords"]
+
+
+@pytest.mark.django_db
+def test_create_new_job_form_with_invalid_data(example_recruiter, example_job_data_as_dictionary,
+                                               client):
+    example_job_data_as_dictionary['title'] = ""  # invalid title
+    client.force_login(example_recruiter.user)
+    response = client.post("/recruiter/create_new_job_form/", data=example_job_data_as_dictionary)
+    assert response.status_code == 200
```
guy9050: Do you get status code 200?

recruiter/views.py
```diff
@@ -73,3 +56,19 @@ def recruiterRegister(request):
 
 def recruiter_created_successfully(request):
     return render(request, 'recruiter/recruiter_created_successfully.html')
+
+
+def create_new_job_form(request):
+    recruiter_object = get_object_or_404(Recruiter, user_id=request.user.id)
```
guy9050: if it returns 404 your functions continues but then will fail/insert invalid data on lines 68-69

guy9050 (commented):
I added some comments, otherwise looks good

SJMaster/urls.py
benmenachem: typo

jobboard/tests.py
benmenachem: will do

jobboard/tests.py
benmenachem: Yea I do get status code 200 since the data invalid so I'm staying at the same page. I believe that's what you spoke about with Barak yesterday in the lecture

recruiter/views.py
benmenachem: If I understand correctly of what I read online, if the user isn't found then get_object_or_404 will raise a 404 error and the function will stop and won't continue

ifireball (changes requested):

recruiter/views.py
ifireball: Yes the 404 error is thrown as an exception so the function does not continue.

jobboard/tests.py
benmenachem: @ifireball I updated the test to check that the form is displayed to the end user 
regarding test for "simple empty form display" I'm not sure I understand what you mean. do you mean you want me to test that the form is empty when the page is loaded?

recruiter/views.py
```diff
@@ -74,4 +57,22 @@ def recruiterRegister(request):
 
 
 def recruiter_created_successfully(request):
+    jobboard.views.add_navbar_links_to_context(request,)
     return render(request, 'recruiter/recruiter_created_successfully.html')
+
+
+def create_new_job_form(request):
```
ifireball: Why have this in `recruiter.views` and not in `jobbard.views`?

ifireball (changes requested):

recruiter/views.py
benmenachem: @ifireball since it's a feature that only a recruiter has I thought it would make more sense to put it in the recruiter.views

ifireball: The models, the forms that edit them, the views that show them and the tests for them typically all go in the same app. Where would you intuitively look for the view that creates new Job objects?

benmenachem: I guess your right. to be honest, intuitively it made sense to me to put it in the recruiter views since it is a recruiter feature. But now I understand your point and it does make more sense to me to put it in the jobboard views. I will fix it now and update the PR.

benmenachem: @ifireball fixed it and moved the function and the form to the jobboard app

jobboard/tests.py
benmenachem: added the requested test

recruiter/views.py
```diff
@@ -71,4 +48,5 @@ def recruiterRegister(request):
 
 
 def recruiter_created_successfully(request):
+    jobboard.views.add_navbar_links_to_context(request, )
```
guyy13: This does nothing as the function needs to get a dictionary as well.
Anyway, I already have a PR open about this so you can remove this line.

benmenachem: fixed it

guyy13 (approved):

jobboard/tests/model_function_tests.py
```diff
@@ -112,3 +116,89 @@ def test_get_jobs_by_major():
...
+def test_navbar_links_are_correct_for_student_user(client, example_student_a):
+    client.force_login(example_student_a.user)
+    response = client.get("/")
+    assert response.status_code == 200
+    assert response.context['navbar_links'] == {"Account Settings": "/student/account_settings/", "Logout": "/logout",
+                                                f"Welcome {example_student_a.user.username}": "#"}
+
+
+@pytest.mark.django_db
+def test_navbar_links_are_correct_for_recruiter_user(client):
```
guyy13: There's a separate file fo the nav_bar tests, not sure why you're adding all of these tests again.
I don't think you need to add anything to the model_function_tests file, only to the view_tests

jobboard/tests/conftest.py
```diff
@@ -21,3 +24,27 @@ def example_student_a(example_institution_a):
                       major=Major.COMPUTER_SCIENCE, about="testing", graduation_date=date(2020, 7, 13))
     student.save()
     return student
+
+
+@pytest.fixture
+def example_recruiter():
```
guyy13, line 30: @ifireball said to keep the conftests file small, so maybe if these fixtures are only being used in the view_tests.py file it's better to put them in the file itself and not here

guyy13 (changes requested):
left some comments, besides that lgtm

jobboard/tests/conftest.py
benmenachem: I saw his comment on Eldar's PR but I think it makes more sense to put the fixtures here since it's just 2 more and we're probably not gonna put more fixtures in that file.

jobboard/tests/model_function_tests.py
benmenachem: fixed it.

jobboard/tests/model_function_tests.py
```diff
@@ -3,6 +3,7 @@
 from student.models import Major
 from datetime import date
 
+
```
guyy13: Can you remove this? No need to change this file if you don't add anything to it, it might cause unnecessary conflicts with different PRs

guyy13 (approved):

guy9050 (approved):
lgtm

PR #72, opened 2021-12-18
Add navbar links

guyy13:
Added navbar links to all of the relevant pages.
Moved the logic out of the base.html template, all of the logic now resides in the function `add_navbar_links_to_context` in 'jobboard/views.py' .
Also included tests for the function to make sure the correct links are showing up.

### DIFF
--- a/SJMaster/templates/base.html
+++ b/SJMaster/templates/base.html
@@ -1,69 +1,57 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
+<head>
     <!-- Required meta tags -->
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1"/>
     <!-- Bootstrap CSS -->
     <link
-      href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css"
-      rel="stylesheet"
-      integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3"
-      crossorigin="anonymous"
+            href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css"
+            rel="stylesheet"
+            integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3"
+            crossorigin="anonymous"
     />
     <title>SJMaster</title>
-  </head>
+</head>
 
-  <body>
-    <header class="site-header">
-      <nav class="navbar navbar-expand-lg navbar-light bg-light">
+<body>
+<header class="site-header">
+    <nav class="navbar navbar-expand-lg navbar-light bg-light">
         <a class="navbar-brand" href="/" style="margin-left: 3em">SJMaster</a>
         <button
-          class="navbar-toggler"
-          type="button"
-          data-toggle="collapse"
-          data-target="#navbarSupportedContent"
-          aria-controls="navbarSupportedContent"
-          aria-expanded="false"
-          aria-label="Toggle navigation"
+                class="navbar-toggler"
+                type="button"
+                data-toggle="collapse"
+                data-target="#navbarSupportedContent"
+                aria-controls="navbarSupportedContent"
+                aria-expanded="false"
+                aria-label="Toggle navigation"
         >
-          <span class="navbar-toggler-icon"></span>
+            <span class="navbar-toggler-icon"></span>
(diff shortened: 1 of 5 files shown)


benmenachem (approved):

guyy13:
Please note that the link pointing to the "Edit Account" page of a Recruiter is pointing to the new URL that's being updated on #68

jobboard/views.py
```diff
@@ -44,4 +45,20 @@ def board(request):
...
+                                   f"Welcome {request.user.username}": "#"}
+
+    elif Recruiter.is_recruiter(request.user.id):
+        context['navbar_links'] = {"Account Settings": "/recruiter/account_settings/", "My Jobs": "/myjobs",
+                                   "Create Job": "/recruiter/create_new_job_form", "Logout": "/logout",
+                                   f"Welcome {request.user.username}": "#"}
+    elif request.user.is_authenticated:
+        context['navbar_links'] = {"Logout": "/logout"}
+    else:
+        context['navbar_links'] = {"Login": "/login"}
```
yaelerdr, line 64: Shouldn't we also check the link of 'Job Search'?

SJMaster/templates/base.html
```diff
@@ -1,69 +1,57 @@
...
-            </li>
-            {% else %}
-            <li class="nav-item">
-              <a class="nav-link" href="/login">Login</a>
-            </li>
-            {% endif %}
-          </ul>
+            <ul class="navbar-nav mr-auto">
+                <li class="nav-item active">
+                    <a class="nav-link" href="/">Job Search</a>
```
yaelerdr, line 35: We might want to enable this tag only when the client isn't on the main page

yaelerdr (changes requested):
I think that the 'Job Search' tag of the nav-bar should only appear when the client isn't on the main page - which is the job search itself (just to not let him clicking an address that takes him to the same address he's in now..)

jobboard/views.py
guyy13: I think it makes sense the the job search item is always visible to everyone as this is the main feature of the website

SJMaster/templates/base.html
guyy13: see my comment above

YarinBou:
lgtm

YarinBou (approved):

yaelerdr (approved):
👍👍👍

jobboard/tests.py
```diff
@@ -182,3 +182,41 @@ def test_page_display_to_user_that_is_recruiter(client):
     jobs_to_display = Job.get_jobs_posted_on_or_after_specific_date(date_six_months_ago)
     # test that jobs displayed are only jobs created 6 months ago
     assert response.context['jobs_to_display'] == jobs_to_display
+
+
+@pytest.mark.django_db
+def test_navbar_links_are_correct_for_student_user(client, example_student_a):
+    client.force_login(example_student_a.user)
```
guy9050, line 189: why `force_login`?

jobboard/tests.py
```diff
@@ -182,3 +182,41 @@ def test_page_display_to_user_that_is_recruiter(client):
...
+def test_navbar_links_are_correct_for_not_logged_in_user(client):
+    response = client.get("/")
+    assert response.status_code == 200
+    assert response.context['navbar_links'] == {"Login": "/login"}
+
+
+@pytest.mark.django_db
+def test_navbar_links_are_correct_for_logged_in_user_not_recruiter_or_student(client):
+    test_user = User.objects.create_user(username="testuser1", password="testuser1")
+    test_user.save()
```
guy9050, line 218: why are you saving the test_user in the DB?

student/views.py
```diff
@@ -1,21 +1,22 @@
...
     return render(request, 'account_settings_update_success.html', context)
 
 
 def update_student_account_settings_view(request):
     context = {}
     student_object = get_object_or_404(Student, user_id=request.user.id)
-    update_student_form = UpdateStudentAccountSettingsForm(
-        request.POST or None, instance=student_object)
-    add_is_student_to_context(request, context)
+    update_student_form = UpdateStudentAccountSettingsForm(request.POST or None, instance=student_object)
```
guy9050, line 18: Can you please explain this line?

jobboard/tests.py
guyy13: @guy9050 why not? It's easier to use instead of typing in the username and password for the user and we used it in plenty of tests already

jobboard/tests.py
guyy13: @guy9050 
Each test creates a temporary DB file on which it is running, thus saving an instance in the DB inside a test doesn't actually save it in the DB (only in the temp DB it's using for the test).

student/views.py
guyy13: @guy9050 
It makes the form handle both a POST request and a GET request, I can change it if you want but this code is already merged and working and not really related to this PR. I just deleted some spaces around it this is why it looks like it's new.

ifireball (approved):

PR #74, opened 2021-12-18
Adding some more example companies.

YarinBou:
Small PR just for adding some basic companies.

### DIFF
--- a/recruiter/tests.py
+++ b/recruiter/tests.py
@@ -60,9 +60,6 @@ def test_get_all_companies_returns_all_companies_as_a_query_set():
     companies_set = Company.get_all_companies()
     assert isinstance(companies_set, QuerySet)
     assert all(isinstance(company, Company) for company in companies_set)
-    assert list(companies_set.values_list("name", "description", "website_url")) == [
-        ("example_company_a", "This is company a", "company_a.com"),
-        ("example_company_b", "This is company b", "company_b.com")]
 
 
 @pytest.mark.django_db
@@ -134,7 +131,8 @@ def test_signup_form_for_recruiter_loads_correctly(valid_recruiter_data, client)
     form = response.context["form"]
     form_initial_data = response.context["form"].initial
     assert isinstance(form, RecuiterRegistrationForm)
-    assert all(form_initial_data[key] == valid_recruiter_data[key] for key in form_initial_data)
+    assert all(form_initial_data[key] == valid_recruiter_data[key]
+               for key in form_initial_data)
 
 
 @pytest.mark.django_db
--- a/recruiter/tests.py
+++ b/recruiter/tests.py
@@ -60,6 +60,17 @@ def test_get_all_companies_returns_all_companies_as_a_query_set():
     companies_set = Company.get_all_companies()
     assert isinstance(companies_set, QuerySet)
     assert all(isinstance(company, Company) for company in companies_set)
+    assert list(companies_set.values_list("name", "description", "website_url")) == [
+        ("Check Point Software Technologies, Ltd.",
+         "This is Check Point Software Technologies, Ltd.",
+         "Check_Point_Software_Technologies.com"),
+        ("Kaplan Open Source", "This is Kaplan Open Source", "KaplanOpenSource.com"),
+        ("Ping Identity", "This is Ping Identity", "PingIdentity.com"),
+        ("Seeking Alpha", "This is Seeking Alpha", "SeekingAlpha.com"),
+        ("Synamedia", "This is Synamedia", "Synamedia.com"),
+        ("Toptal", "This is Toptal", "Toptal.com"),
+        ("example_company_a", "This is company a", "company_a.com"),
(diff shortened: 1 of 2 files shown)


guyy13 (changes requested):
Based on previous reviews from the mentors, I don't think you should an existing migration file and instead create a new one to add the data.

YarinBou:
if you're making a new migration that's spouse to be fine, what's wrong with that?

guyy13:
> if you're making a new migration that's spouse to be fine, what's wrong with that?

I don't understand what you're saying.
I'm just saying, base on previous reviews from @ifireball , we should not change merged migration files, so instead just create a new migration file with the new data.

YarinBou:
I just asked what's the difference?

guyy13:
> I just asked what's the difference?

According to Barak it can break staging systems. Can't say I fully understand what it means, but you can look at the comments here, I'm just trying to save us time
https://github.com/redhat-beyond/SJMaster/pull/47

benmenachem (approved):

jobboard/migrations/0008_merge_20211219_1816.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('jobboard', '0004_alter_job_work_from'),
+        ('jobboard', '0007_add_example_companies'),
+    ]
+
+    operations = [
```
guyy13: What's the point of this file? Looks like it doesn't do anything or am I missing something?

guyy13 (changes requested):
Besides that one comment lgtm

jobboard/migrations/0008_merge_20211219_1816.py
YarinBou: when I added a new migration file for some reason I had a conflict between 004 and 007 files, In this way, I didn't change them and resolved the conflict with the Django migration default option.

guyy13: Ok, looks a bit weird that the operations list is empty, but maybe it's alright

guyy13 (approved):

eldaraharonson (commented):
I think we should also add jobs for these companies, are you planning to do that in a another PR @YarinBou ?

YarinBou:
> I think we should also add jobs for these companies, are you planning to do that in a another PR @YarinBou ?

you can do another PR for that

jobboard/migrations/0007_add_example_companies.py
```diff
@@ -0,0 +1,36 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('jobboard', '0001_initial'),
+    ]
+
+    def generate_data(apps, schema_editor):
+
+        with transaction.atomic():
```
ifireball: The Company model belongs to the `recruiter` app, so why does this migration in the `jobboard` app? the end result may be that the migration order is messed up since we do not define the right dependencies.

ifireball (changes requested):

jobboard/migrations/0007_add_example_companies.py
YarinBou: I do that because our other example companies also were there, so I think that that is their place. I will add the migrations over there, is that ok to delete these files in my local computer and add the new file on recruiter?  that will fix the right dependencies back?

guyy13: Save the code that you want to add somewhere on your computer,
then on your local branch, restore all the migration files to how they are on upstream/main.
Then create a new migration file in the correct location with the code you saved.
you can use `git restore` in order to restore a file to how it is on the main repo.

YarinBou: hey, guy thx for your help! that's worked. 
I upload the updated version. Good night.

recruiter/tests.py
```diff
@@ -60,9 +60,6 @@ def test_get_all_companies_returns_all_companies_as_a_query_set():
     companies_set = Company.get_all_companies()
     assert isinstance(companies_set, QuerySet)
     assert all(isinstance(company, Company) for company in companies_set)
```
ifireball, line 62: If we drop the assertions below, how do we know that we're getting the valid company data and not some garbage?

ifireball (changes requested):

recruiter/tests.py
YarinBou: It just seems to me a bit "dirty" code to check each name one by one as it was

YarinBou: We check if we get back an company instance's

ifireball: What guarantee do you have that the instance fields contain proper data? Maybe a string decoder got messed up somewhere?
Also what about the order of responses? You were checking that as well with the comparison before.

YarinBou: ok, done.

ifireball (approved):

PR #79, opened 2021-12-23
Search Bar Feature

eldaraharonson:
The search bar is built of 2 different bars:
  1. On the first bar the user can input a keyword corresponding to a
     job or the name of a company.
  2. On the second bar the user can input a city name.

The user can decide if he wants to search only by keyword/company,
only by city, or by both keyword/company and city. If the user chose to
search by both the the jobs that will be displayed will be only jobs
that match both search bars.

Example: user inputs on first bar 'software' and on second bar 'tel
aviv-yafo' then only jobs that have a software keyword **and** their location
is tel aviv-yafo will be displayed.

image

### DIFF
--- a/jobboard/templates/jobboard/board.html
+++ b/jobboard/templates/jobboard/board.html
@@ -1,48 +1,70 @@
 {% extends 'base.html' %} {% block content %}
 
 <div class="container-fluid">
-    <div class="row">
-        <div class="col-2">
-            <div class="user-info">
-                <div class="container-fluid">
-                    {% if user_is_student or user_is_recruiter %}
-                        <h2>{{user.first_name}} {{user.last_name}}</h2>
-                    {% if user_is_student %}
-                        <h4>{{user_as_student.educational_institution}}</h4>
-                    {% endif %}
-                    {% if user_is_recruiter %}
-                        <h4>{{user_as_recruiter.company}}</h4>
-                    {% endif %} {% endif %}
-                </div>
-            </div>
-        </div>
-        <div class="col-10 text-center">
-            <div class="row">
-                {% if user_is_student %}
-                    <h3>Here are some jobs that match your profile</h3>
-                {% else %}
-                    <h3>Here are jobs created in the last 6 months</h3>
-                {% endif %}
-                {% for job in jobs_to_display %}
-                    <div class="col-6 p-5 text-center">
-                        <div class="jobs">
-                            <ul class="list-group">
-                                <li class="list-group-item">{{job.title}}</li>
-                                <li class="list-group-item">Added {{job.date_created}}</li>
-                                <li class="list-group-item">{{job.company}}</li>
-                                <li class="list-group-item">{{job.job_type}}</li>
-                                <li class="list-group-item">{{job.city}}</li>
-                                <li class="list-group-item">
-                                <button class="list-group-item" onclick="location.href ='/job/{{ job.id }}/'">Read more
-                                </button>
-                                </li>
-                            </ul>
-                        </div>
-                    </div>
-                {% endfor %}
-            </div>
-        </div>
+  <div class="row">
+    <div class="col-4"></div>
+    <div class="col-8">
+      <form class="d-flex my-3" method="POST" action="{% url 'board' %}">
+        {% csrf_token %}
+        <input
+          class="form-control me-2"
+          type="search"
+          placeholder="Search company name, job keyword"
+          aria-label="Search"
+          name="searched_name"
+        />
+        <input
+          class="form-control me-2"
+          type="search"
+          placeholder="Search city"
+          aria-label="Search"
+          name="searched_city"
+        />
+        <button class="btn btn-outline-success" type="submit">Search</button>
+      </form>
     </div>
-</div>
+  </div>
+  <div class="row">
+    <div class="col-2">
+      <div class="user-info">
+        <div class="container-fluid">
+          <h2>{{user_name}}</h2>
+          {% if user_is_student %}
+            <h4>{{user_as_student.educational_institution}}</h4>
+          {% endif %}
+          {% if user_is_recruiter %}
+            <h4>{{user_as_recruiter.company}}</h4>
+          {% endif %}
+        </div>
+      </div>
+    </div>
+    <div class="col-10 text-center">
+      <div class="row">
+        <h3>{{jobs_intro_message}}</h3>
+        {% for job in jobs_to_display %}
+          <div class="col-6 p-5 text-center">
+            <div class="jobs">
+              <ul class="list-group">
+                <li class="list-group-item">{{job.title}}</li>
+                <li class="list-group-item">Added {{job.date_created}}</li>
+                <li class="list-group-item">{{job.company}}</li>
+                <li class="list-group-item">{{job.job_type}}</li>
+                <li class="list-group-item">{{job.city}}</li>
+                <li class="list-group-item">
+                  <button
+                    class="list-group-item"
+                    onclick="location.href ='/job/{{ job.id }}/'"
+                  >
+                    Read more
+                  </button>
+                </li>
+              </ul>
+            </div>
+          </div>
+        {% endfor %}
+      </div>
+    </div>
 
-{% endblock %}
+    {% endblock %}
+  </div>
+</div>
--- a/jobboard/tests/search_bar_tests.py
+++ b/jobboard/tests/search_bar_tests.py
@@ -1,0 +1,98 @@
+import pytest
+from pytest_django.asserts import assertTemplateUsed
+from jobboard.models import Job
+
+# The tests use the data that is already in the db from the migration files
+
+
+@pytest.mark.django_db
+def test_search_bar_empty_search(client, example_student_a):
+    client.force_login(example_student_a.user)
+    search_bar_data = {"searched_name": "", "searched_city": ""}
+    response = client.post("/", data=search_bar_data)
+    assert response.status_code == 200
+    assertTemplateUsed(response, 'jobboard/board.html')
+    assert response.context["jobs_to_display"] == set()
+
+
+@pytest.mark.django_db
+def test_search_bar_company(client, example_student_a):
+    job_b = Job.objects.get(title="job_b")
+    job_d = Job.objects.get(title="job_d")
+    client.force_login(example_student_a.user)
+    # searches for jobs at 'example_company_b'
+    search_bar_data = {"searched_name": "Example_Company_B", "searched_city": ""}
+    response = client.post("/", data=search_bar_data)
+    assert response.status_code == 200
+    assertTemplateUsed(response, 'jobboard/board.html')
+    assert response.context["jobs_to_display"] == {job_b, job_d}
+
+
+@pytest.mark.django_db
+def test_search_bar_keyword(client, example_student_a):
+    job_a = Job.objects.get(title="job_a")
+    job_b = Job.objects.get(title="job_b")
+    job_e = Job.objects.get(title="job_e")
+    client.force_login(example_student_a.user)
+    # searches for jobs with keyword 'software'
+    search_bar_data = {"searched_name": "software", "searched_city": ""}
+    response = client.post("/", data=search_bar_data)
+    assert response.status_code == 200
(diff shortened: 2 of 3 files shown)


SJMaster/templates/base.html
```diff
@@ -1,57 +1,69 @@
...
-            <span class="navbar-toggler-icon"></span>
+          <span class="navbar-toggler-icon"></span>
         </button>
+
         <div class="collapse navbar-collapse" id="navbarSupportedContent">
-            <ul class="navbar-nav mr-auto">
-                <li class="nav-item active">
-                    <a class="nav-link" href="/">Job Search</a>
-                </li>
-                {% for link_title, link_address in navbar_links.items %}
```
guyy13: I think you didn't rebase your branch properly, you shouldn't delete this.

jobboard/models.py
```diff
@@ -47,11 +47,20 @@ class Job(models.Model):
...
 
     @classmethod
-    def get_jobs_by_city_name(cls, city_name):
+    def get_jobs_by_city_name(cls, city_names_as_string):
         """
         Allows students to search for a job based on the city it is located in
         """
-        return set(cls.objects.filter(city=City.objects.get(name=city_name)))
+        jobs_to_return = set()
+        city_names = city_names_as_string.split()
```
guyy13: I think splitting the string like this might cause issues if the city name is not one word like "Tel Aviv" for example.

guyy13:
Since this PR is quite big, maybe put the model functions in a different PR?

jobboard/models.py
```diff
@@ -98,11 +110,20 @@ def get_jobs_by_work_model(cls, work_model):
...
     @classmethod
-    def get_jobs_by_company_name(cls, company_name):
+    def get_jobs_by_company_name(cls, companies_names_as_string):
         """
-        Allows students to search for a job based on the company
+        Allows students to search for a job based on the companies
         """
-        return set(cls.objects.filter(company=Company.objects.get(name=company_name)))
+        jobs_to_return = set()
+        company_names = companies_names_as_string.split()
```
guyy13: Similar issue with the split on the cities function, a company like "Red Hat" would split into to different companies "Red" and "Hat"

jobboard/templates/jobboard/board.html
```diff
@@ -17,7 +40,13 @@ <h4>{{user_as_recruiter.company}}</h4>
     </div>
     <div class="col-10 text-center">
       <div class="row">
-        {% if user_is_student %}
+        {% if searched_name and not searched_area %}
```
guyy13: Maybe add a "title" key to the context dictionary, than you can handle all these ifs outside of the template

guyy13 (changes requested):
left some comments, good job Eldar.

jobboard/models.py
```diff
@@ -72,15 +81,18 @@ def get_jobs_by_job_type(cls, job_type):
         return set(cls.objects.filter(job_type=job_type))
 
     @classmethod
-    def get_jobs_by_keywords(cls, *keywords_as_string):
+    def get_jobs_by_keywords(cls, keywords_as_string):
         """
-        Allows students to search for a job based on one ore more keywords linked to the job
+        Allows students to search for a job based on one ore more keywords linked to the job.
```
yaelerdr: or* :)

yaelerdr: and - why deleting the * ?
Wasn't it there so we could send this fun multiple keywords?

jobboard/views.py
```diff
@@ -38,13 +37,38 @@ def get_content_if_user_is_not_student(request, user_is_recruiter):
...
+        content["searched_area"] = searched_area
+        if searched_name and not searched_area:
+            content['jobs_to_display'] = searched_by_companies_or_keywords(searched_name)
+        elif searched_area and not searched_name:
+            content['jobs_to_display'] = searched_by_area(searched_area)
+        else:
+            content['jobs_to_display'] = searched_by_area(searched_area).\
+                intersection(searched_by_companies_or_keywords(searched_name))
     add_navbar_links_to_context(request, content)
     return render(request, 'jobboard/board.html', content)
```
yaelerdr, line 83: I think we should consider the case in which non-job corresponds to the search. Can maybe be implemented by  initializing jobs_to_display to None, then doing the same logic you have implemented with the "if.. eleif..", and then to add something like 
`if  content['jobs_to_display'] is None:
      return render(request, 'non_job_matched_the_search.html') `
 and also add that "non_job_matched_the_search.html" page

jobboard/models.py
yaelerdr: Yeah, I agree with @guyy13 . I think we should leave that function the way it was.
If for example someone search Tel Aviv, Tel Mond jobs could also be returned out of that func

jobboard/templates/jobboard/board.html
```diff
@@ -17,7 +40,13 @@ <h4>{{user_as_recruiter.company}}</h4>
...
+        <h3>Jobs that match "{{searched_name}}"</h3>
+        {% elif searched_area and not searched_name %}
+        <h3>Jobs that match "{{searched_area}}"</h3>
+        {% elif searched_area and searched_name %}
+        <h3>Jobs that match "{{searched_name}} and "{{searched_area}}"</h3>
+        {% elif user_is_student %}
         <h3>Here are some jobs that match your profile</h3>
         {% else %}
         <h3>Here are jobs created in the last 6 months</h3>
```
yaelerdr: Adding indentation to the if s could really help understand the code.. Happened to me too automatically. Not sure why.. Fixed it manually :\

ifireball:
This is a huge PR and a huge piece of functionality, please try to split it apart - put refactoring steps in a separate PR and make this one only focus on new functionality.

jobboard/models.py
guyy13: @yaelerdr 
That's a bug, if I search for Tel Aviv I want to see jobs in Tel Aviv, not in Tel Mond.

eldaraharonson: I agree this can be a bit problematic however I can't think of a better way to implement this. The user won't necessarily type "Tel-Aviv" in the search bar and if he types "Tel Aviv" then the way you guys are suggesting won't return any of the Tel-Aviv jobs. I think displaying jobs that the user doesn't specifically search for (like Tel Mond when searched for Tel-Aviv)  in addition to the ones he did search for is better then displaying 0 jobs because he didn't add a "-". Also if I return the model function to the way it was before, the user can only search 1 city at a time and I think its good that he can search a bunch of cities at once. 
But I'm open to suggestions if you guys can think of a better way to implement this.

eldaraharonson:
> This is a huge PR and a huge piece of functionality, please try to split it apart - put refactoring steps in a separate PR and make this one only focus on new functionality.

Will do

jobboard/templates/jobboard/board.html
eldaraharonson: You're right, for me it's because I use the "prettier" extension in vscode, it reformats html code automatically when saving but it doesn't format Django code additions nicely. I will fix it.

jobboard/models.py
eldaraharonson: Because this function is called with the input entered in the search bar therefore the argument won't be an iterable. I use the .split() function on the argument so that if the user enters more than one keywords it puts them in a list and then I can still search for jobs based on each keyword.

jobboard/views.py
eldaraharonson: I don't know if another html page is necessary but I can display a text saying "No jobs matched this search" or something like that.

SJMaster/templates/base.html
eldaraharonson: I'll readd it

jobboard/tests/view_tests.py
```diff
@@ -0,0 +1,76 @@
+import pytest
```
YarinBou: that's a good idea to split the tests.

jobboard/views.py
```diff
@@ -38,13 +37,38 @@ def get_content_if_user_is_not_student(request, user_is_recruiter):
     return content
 
 
+def searched_by_companies_or_keywords(searched_name):
+    jobs_by_company = Job.get_jobs_by_company_name(searched_name)
+    jobs_by_keywords = Job.get_jobs_by_keywords(searched_name)
+    return jobs_by_company.union(jobs_by_keywords)
+
+
+def searched_by_area(searched_area):
```
YarinBou: why not search by city? if that what we do here? or if that was a region so search by region?

YarinBou (approved):

ifireball:
I don't understand why I need to comment on this kind of thing at this stage, but please move the splitting of the tests file to a different PR so it becomes possible to find the new tests added for the new functionality.

Also given the time and comment for team members, I would consider trying to add the gist of this functionality without trying to go for for free-text parsing.

jobboard/views.py
eldaraharonson: You're right, I'll change it

jobboard/models.py
yaelerdr: @eldaraharonson We can maybe add the function of a complete match to a city in DB, run this one according to the user's search, and if there wasn't a perfect match, then run the function of partially matched results.

jobboard/templates/jobboard/board.html
eldaraharonson: Done

jobboard/views.py
benmenachem: @eldaraharonson so what happens eventually if no matches are found ? does it display that no jobs matched the search? cus I only see 3 cases here

yaelerdr (approved):

SJMaster/templates/base.html
```diff
@@ -61,4 +61,4 @@
         crossorigin="anonymous"
 ></script>
 </body>
-</html>
+</html>
```
guyy13: Seems a bit pointless can you fix it?

jobboard/tests/search_bar_tests.py
```diff
@@ -0,0 +1,108 @@
...
+def test_search_bar_company(client, example_student_a):
+    job_a = Job.objects.get(title="job_a")
+    job_b = Job.objects.get(title="job_b")
+    job_c = Job.objects.get(title="job_c")
+    job_d = Job.objects.get(title="job_d")
+    job_e = Job.objects.get(title="job_e")
+    client.force_login(example_student_a.user)
+    # searches for jobs at 'example_company_b'
+    search_bar_data = {"searched_name": "Example_Company_B", "searched_city": ""}
+    response = client.post("/", data=search_bar_data)
```
guyy13, line 25: Also, check the response status code and template used (needs to be added to all of the tests).

jobboard/tests/search_bar_tests.py
```diff
@@ -0,0 +1,108 @@
...
+    job_d = Job.objects.get(title="job_d")
+    job_e = Job.objects.get(title="job_e")
+    client.force_login(example_student_a.user)
+    # searches for jobs with keyword 'software'
+    search_bar_data = {"searched_name": "software", "searched_city": ""}
+    response = client.post("/", data=search_bar_data)
+    assert response.context["jobs_to_display"] == {job_a, job_b, job_e}
+    # searches for jobs with keyword 'product'
+    search_bar_data = {"searched_name": "product", "searched_city": ""}
+    response = client.post("/", data=search_bar_data)
```
guyy13: I remember Barak said that we should try to keep each test with one HTTP request only. So maybe testing a search with only one keyword is enough. (Same goes for the other tests).

guyy13 (changes requested):
Left a few comments, besides that lgtm.

YarinBou (approved):

jobboard/tests/search_bar_tests.py
eldaraharonson: Done

jobboard/tests/search_bar_tests.py
eldaraharonson: Done

SJMaster/templates/base.html
eldaraharonson: Fixed

guyy13 (approved):
lgtm, goob job Eldar.

benmenachem (approved):

guy9050 (approved):

PR #80, opened 2021-12-24
Split the jobboard tests to separate files

eldaraharonson:
I split the jobboard tests to 3 separate files according to categories.

### DIFF
--- a/jobboard/tests/conftest.py
+++ b/jobboard/tests/conftest.py
@@ -1,0 +1,23 @@
+import pytest
+from student.models import Major
+from datetime import date
+from django.contrib.auth.models import User
+from student.models import Student, EducationalInstitution
+
+
+@pytest.fixture
+def example_institution_a():
+    institution = EducationalInstitution(name="A")
+    institution.save()
+    return institution
+
+
+@pytest.fixture
+def example_student_a(example_institution_a):
+    example_user = User.objects.create_user("test", "testpassword")
+    example_user.save()
+    student = Student(user=example_user, full_name="test user", email="test@test.com", date_of_birth=date(2020, 7, 13),
+                      phone_number="0000000000", educational_institution=example_institution_a,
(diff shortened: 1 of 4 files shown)


jobboard/tests/view_tests.py
```diff
@@ -0,0 +1,76 @@
...
+from datetime import date, timedelta
+from django.contrib.auth.models import User
+from student.models import Student, EducationalInstitution
+from recruiter.models import Recruiter
+
+# The tests use the data that is already in the db from the migration files
+
+
+@pytest.fixture
+def example_institution_a():
```
guyy13: Both files use the exact same fixtures, maybe you can make one file import the fixtures from the other to avoid code duplication?

guyy13 (commented):
left one comment, besides that lgtm

jobboard/tests/view_tests.py
eldaraharonson: Done

guyy13 (approved):
lgtm

benmenachem (approved):
looks good 
notice that PR #69 adds more tests to the jobboard so in case it will be merged before this PR it will cause conflicts

eldaraharonson:
> looks good notice that PR #69 adds more tests to the jobboard so in case it will be merged before this PR it will cause conflicts

If it is merged before I will make the necessary adjustments

jobboard/tests/conftest.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+from student.models import Major
+from datetime import date
+from django.contrib.auth.models import User
+from student.models import Student, EducationalInstitution
```
ifireball, line 5: While `conftest.py` files are a popular feature of pytest I usually advise teams against using them since they essentially create global implicit dependencies. You end up having a hard time figuring out where the fixtures are coming from.

Its small enough to be acceptable for now, but make sure it stays small!

ifireball (approved):

PR #82, opened 2021-12-26
Student account settings page design

guyy13:
Changed the design of the Student's edit account settings page.
Also created a 'static' folder in order to implement a local CSS file.

More PRs will follow to update all forms on the website to look the same.

Screenshot:
Edit_account_settings

Edit: the form is exactly in the middle of the screen, I'm just bad at taking screenshots

### DIFF
--- a/student/forms.py
+++ b/student/forms.py
@@ -9,6 +9,16 @@ class UpdateStudentAccountSettingsForm(forms.ModelForm):
         model = Student
         fields = ["full_name", "email", "date_of_birth", "phone_number", "educational_institution",
                   "major", "about", "graduation_date"]
+        widgets = {
+            'full_name': forms.TextInput(attrs={'class': 'form-control input-black-text'}),
+            'email': forms.TextInput(attrs={'class': 'form-control input-black-text'}),
+            'date_of_birth': forms.TextInput(attrs={'class': 'form-control input-black-text'}),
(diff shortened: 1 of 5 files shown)


YarinBou (approved):

eldaraharonson (approved):
Nice font!

benmenachem (approved):
lgtm 
so since your editing the base.html I assume all the design PR's should be merged after this one right?

yaelerdr (approved):
looks good!! nice work 👍

student/forms.py
```diff
@@ -9,6 +9,16 @@ class Meta:
         model = Student
         fields = ["full_name", "email", "date_of_birth", "phone_number", "educational_institution",
                   "major", "about", "graduation_date"]
+        widgets = {
```
ifireball, line 12: There are a few issues with this implementation:

1. In general this kind of styling info does not belong in the code like this - its better be done in the template files. Its was a but inconvenient to do that in Django up to version 4.0 so I understand why you didn't.
2. There is also a massive amount of code repetition here you might want to avoid in the future.
3. By manually writing all the widgets you increase your maintenance costs because youll have ti make sure you keep it in sync in the future
4. By implementing things like this just for one particular form, you'll have to keep re-implementing for each and every form.

It might not be simple for you to fix all thosae issues for now, so I will merge the code as-is

ifireball (approved):

PR #84, opened 2021-12-26
Recruiter account settings page design

guyy13:
Changed the design of the Recruiter's edit account settings page.

Screenshot:
rec_account_settings

### DIFF
--- a/recruiter/forms.py
+++ b/recruiter/forms.py
@@ -2,6 +2,7 @@ from django import forms
 from django.contrib.auth.forms import UserCreationForm
 from django.contrib.auth.models import User
 from .models import Recruiter, Company
+from jobboard.forms import FormControl
 
 
 class RecuiterRegistrationForm(UserCreationForm):
@@ -29,7 +30,7 @@ class RecuiterRegistrationForm(UserCreationForm):
         return user2
 
 
-class UpdateRecruiterAccountSettingsForm(forms.ModelForm):
+class UpdateRecruiterAccountSettingsForm(FormControl):
     class Meta:
         model = Recruiter
         fields = ["name", "company", "email", "phone_number"]
--- a/jobboard/forms.py
+++ b/jobboard/forms.py
@@ -6,3 +6,10 @@ class CreateNewJobForm(forms.ModelForm):
     class Meta:
         model = Job
         fields = ['title', 'job_type', 'major', 'work_from', 'description', 'city', 'address', 'title_keywords']
+
+
+class FormControl(forms.ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        for form_field in self.visible_fields():
+            form_field.field.widget.attrs['class'] = 'form-control'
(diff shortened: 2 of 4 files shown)


YarinBou (approved):

eldaraharonson (approved):

benmenachem (approved):

yaelerdr (approved):

recruiter/forms.py
```diff
@@ -33,3 +33,9 @@ class UpdateRecruiterAccountSettingsForm(forms.ModelForm):
     class Meta:
         model = Recruiter
         fields = ["name", "company", "email", "phone_number"]
+
+    # Override the init method to add classes to the form fields
+    def __init__(self, *args, **kwargs):
+        super(UpdateRecruiterAccountSettingsForm, self).__init__(*args, **kwargs)
+        for form_field in self.visible_fields():
+            form_field.field.widget.attrs['class'] = 'form-control'
```
ifireball: Nice, but how would you avoid having to repeat this code in all the forms?

guyy13: I'll create a function that does that and then will call it from each `init` method.
Will push it shortly

guyy13: @ifireball 
Done,
I wasn't sure where to put this function so I put it in `jobboard/forms.py` as `jobboard` is sort of the main app, so it made sense to me.
Please let me know if it makes more sense to put it somewhere else.

recruiter/forms.py
```diff
@@ -33,3 +34,8 @@ class UpdateRecruiterAccountSettingsForm(forms.ModelForm):
     class Meta:
         model = Recruiter
         fields = ["name", "company", "email", "phone_number"]
+
+    # Override the init method to add classes to the form fields
+    def __init__(self, *args, **kwargs):
+        super(UpdateRecruiterAccountSettingsForm, self).__init__(*args, **kwargs)
```
ifireball: You don't need to pass argument to `super()` in Python 3. Unless you need precise control over the superclass selection.

recruiter/forms.py
```diff
@@ -33,3 +34,8 @@ class UpdateRecruiterAccountSettingsForm(forms.ModelForm):
     class Meta:
         model = Recruiter
         fields = ["name", "company", "email", "phone_number"]
+
+    # Override the init method to add classes to the form fields
```
ifireball: you'd still need to write this kind of boilerplate code for all form object in your system - a batter approach would be to subclass ModelForm, add your customization and then make all your form classes inherit from that.

You can use multiple inheritance in Python, so this could probably be used just fine for classes that do not inherit directly from ModelForm as well.

guyy13: Let me just make sure I follow,
So I should create a class that inherits from ModelForm, overwrites the `init` method, and then make all the forms inherit from that class?

recruiter/forms.py
guyy13: Fixed.

recruiter/forms.py
guyy13: @ifireball Ok I think I managed to do it

guyy13:
Fixed conflicts

ifireball (approved):

YarinBou (approved):

benmenachem (approved):

yaelerdr (approved):

eldaraharonson (approved):

guy9050 (approved):

PR #86, opened 2021-12-27
Changed Jobboard Model Functions for Search Bar

eldaraharonson:
Changed the jobboard model functions that fetch jobs based on a keyword,
city, or company so that only one search value can be sent to each
function. These functions will be used by the search bar that is being implemented.
Updated the tests accordingly.

### DIFF
--- a/jobboard/models.py
+++ b/jobboard/models.py
@@ -51,7 +51,7 @@ class Job(models.Model):
         """
         Allows students to search for a job based on the city it is located in
         """
-        return set(cls.objects.filter(city=City.objects.get(name=city_name)))
+        return set(cls.objects.filter(city__name__iexact=city_name))
 
     @classmethod
     def get_jobs_by_region_name(cls, region_name):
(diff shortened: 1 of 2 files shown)


YarinBou (approved):

jobboard/models.py
```diff
@@ -51,7 +51,11 @@ def get_jobs_by_city_name(cls, city_name):
         """
         Allows students to search for a job based on the city it is located in
         """
-        return set(cls.objects.filter(city=City.objects.get(name=city_name)))
+        try:
```
guyy13: I think you can do all of this function in one line like so:
`return set(cls.objects.filter(city__name__iexact=city_name))`
Same goes for the other functions.

eldaraharonson: Done

guyy13 (approved):
lgtm

yaelerdr (approved):
👍

benmenachem (approved):

ifireball (approved):
Approving for now, but given the complexity of the search bar, I suspect you'll need to add additional tests here to test for the tricky edge cases such as wen a city name has a space character in it.

